PR #105, opened 2015-05-17
Handle arbitrary inputs to BlockWeightedLeastSquares

### LINKED ISSUE #104: Handle arbitrary inputs in block weighted solver
Right now the block weighted least squares solver requires data to be partitioned by classes. We should make it handle arbitrary inputs and do an appropriate shuffle to organize data as required.

shivaram:
Test this with a shuffled version of the same matrices we have. Also add some more scaladoc

Closes #104


src/main/scala/nodes/learning/BlockLinearMapper.scala
```diff
@@ -127,9 +139,22 @@ class BlockLinearMapper(
   }
 }
 
+/**
+ * Fits a least squares model using block co-ordinate descent with provided
```
shivaram: I actually think this is 'coordinate descent' as per wikipedia :) http://en.wikipedia.org/wiki/Coordinate_descent

src/main/scala/nodes/learning/BlockWeightedLeastSquares.scala
```diff
@@ -270,4 +315,39 @@ object BlockWeightedLeastSquaresEstimator extends Logging {
     a
   }
 
+  def groupByClasses(
+      features: Seq[RDD[DenseVector[Double]]],
+      labels: RDD[DenseVector[Double]])
+    : (Seq[RDD[DenseVector[Double]]], RDD[DenseVector[Double]]) = {
+
+    val nClasses = labels.first.length.toInt
+    val hp = new HashPartitioner(nClasses)
```
shivaram: Comment added now

src/main/scala/nodes/learning/BlockWeightedLeastSquares.scala
```diff
@@ -270,4 +315,39 @@ object BlockWeightedLeastSquaresEstimator extends Logging {
...
+        (classIdx, (i * n + k, item))
+      }
+    }.partitionBy(hp).values.mapPartitions { part =>
+      part.toArray.sortBy(_._1).map(_._2).iterator
+    }
+
+    val shuffledFeatures = features.map { featureRDD =>
+      featureRDD.zip(labels).mapPartitionsWithIndex { case (k, iter) =>
+        iter.zipWithIndex.map { case (item, i) =>
+          val classIdx = item._2.toArray.indexOf(item._2.max)
```
shivaram: Done

src/main/scala/nodes/learning/BlockWeightedLeastSquares.scala
```diff
@@ -270,4 +315,39 @@ object BlockWeightedLeastSquaresEstimator extends Logging {
...
+    val nClasses = labels.first.length.toInt
+    val hp = new HashPartitioner(nClasses)
+
+    val n = labels.partitions.length.toLong
+
+    // Associate a unique id with each item
+    // as Spark does not gaurantee two groupBy's will come out in the same order.
+    val shuffledLabels = labels.mapPartitionsWithIndex { case (k, iter) =>
+      iter.zipWithIndex.map { case (item, i) =>
+        val classIdx = item.toArray.indexOf(item.max)
```
shivaram: Ah thanks - didn't know about argmax :)

src/main/scala/nodes/learning/BlockWeightedLeastSquares.scala
```diff
@@ -270,4 +315,39 @@ object BlockWeightedLeastSquaresEstimator extends Logging {
...
+        (classIdx, (i * n + k, item))
+      }
+    }.partitionBy(hp).values.mapPartitions { part =>
+      part.toArray.sortBy(_._1).map(_._2).iterator
```
shivaram: Yeah this part sucks and is because of the way Spark does shuffles right now - So if you have keys from 2 blocks say block1, block2 coming in, they needn't come in with the same order each time. So for example in the first shuffle we might get block1 ++ block2 and in the next one block2 ++ block1

shivaram:
Alright I addressed most of the comments and cleaned up some of the tests as well. Tests are still happy

shivaram:
This is strange - because amplab/keystone was made public, I am no longer able to push updates to my branch. Investigating

etrain:
I had to do a fresh fork on github and push updates from the new repo.

shivaram:
FWIW I just created a github ticket and they managed to synch the fork and the PR ! Ready for more review now...

tomerk:
I had to switch my own fork to public and that fixed it

PR #135, opened 2015-05-22
Removing redundant call to vl_dsift_get_descriptors.

etrain:
Thanks @gylfi for pointing this out (and writing most of this code in the first place :+1: )

This may buy us something like a 2x performance win in SIFT featurization, too.

I have rebuilt and tested locally and the SIFT unit tests still pass.


shivaram:
Wow -- 2x win in SIFTs would be cool ! Could you give a bit of background on what is going on out there ?

shivaram:
Huh I see that it was just a duplicated line of code. Alright. Merging this. LGTM

PR #140, opened 2015-06-01
Pick up Spark, Hadoop versions from environment variables

shivaram:
This PR makes SBT use `SPARK_VERSION` and `SPARK_HADOOP_VERSION` to pick up the dependency versions

cc @etrain

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,23 +12,29 @@ scalaVersion := "2.10.4"
 
 parallelExecution in Test := false
 
+libraryDependencies ++= Seq(
+  "edu.arizona.sista" % "processors" % "3.0" exclude("ch.qos.logback", "logback-classic"),
+  "edu.arizona.sista" % "processors" % "3.0" classifier "models",
+  "org.slf4j" % "slf4j-api" % "1.7.2",
+  "org.slf4j" % "slf4j-log4j12" % "1.7.2",
+  "org.scalatest" %% "scalatest" % "1.9.1" % "test",
+  "org.apache.commons" % "commons-compress" % "1.7",
+  "commons-io" % "commons-io" % "2.4",
+  "org.scalanlp" % "breeze_2.10" % "0.11.2",
+  "com.github.fommil.netlib" % "all" % "1.1.2" pomOnly(),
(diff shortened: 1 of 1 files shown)


etrain:
The point of this patch is so that KeystoneML can support user-specified versions of Spark/Hadoop.

If a user already has a cluster or a local version of Spark installed that is a specific version, KeystoneML should be built against this version. Unfortunately there's no clean way to determine this automatically, so specifying these as environment variables is required.

There is a separate issue of binary incompatibility that comes from our build instructions - we package up the version of spark that is defined in our build file - if you attempt to run with a conflicting version of Hadoop or Spark (via spark-submit) there may be some conflicts at runtime due to conflicting versions of the library being on the classpath. This patch partially alleviates this problem (in that now users can at least tell the system *which* versions they want to build without editing the file), it means that KeystoneML binaries aren't easily transferrable. There are a few different ways to solve this particular problem, but we will save those for a later PR.

The three situations we want to be able to support:
1. Person on their laptop just wants to run something locally (without installing spark): this is supported in the current release and the current patch.
2. Person with spark installed on their laptop wants to run something locally: this is supported currently if you have the *right* version (as specified in the build file) of Spark, Hadoop, etc. installed on your system. This patch makes it work with a user-specified version of Spark/Hadoop.
3. Person wants to spin up a Spark cluster on EC2 and build/run KeystoneML pipelines on it. Again, this works for the *right* version of Spark, Hadoop, etc., but won't necessarily work for an arbitrary version. This patch doesn't (yet) address this situation, but we can set the user-specified environment variables in the keystone-ec2 script and build Keystone on EC2 with the right version numbers.

@shivaram - It looks like a change [here](https://github.com/shivaram/spark-ec2/blob/keystone/keystone/setup.sh#L17) might be what it takes to address 3.

Is this an accurate description of the current state of the build and support for multiple spark versions?

shivaram:
The description is accurate. And FWIW the EC2 scripts already set the Hadoop major version as 2 https://github.com/amplab/keystone/blob/master/bin/keystone-ec2.sh#L9 and our default Hadoop version matches what the spark-ec2 launch with that flag.  We can also set the spark-version in the same manner for the EC2 scripts

etrain:
Great - my comment was more about making sure that SPARK_VERSION gets passed along to `sbt` when keystone is built.

PR #142, opened 2015-06-08
Dag construction

tomerk:
@etrain @shivaram 
Still working on this, but I think it's ready to start getting feedback.

## What is a Pipeline?
A `Pipeline` is made up of a sequence of nodes (where adjacent nodes have matching output/input types). It exposes a ```.fit()``` that outputs a PipelineModel (a transformer made up of a chain of transformers, that applies each one on the input in sequence)

`Pipeline.fit()` is roughly defined as follows:
```scala
def fit(): PipelineModel[A, B] = {
  var transformers: Seq[Transformer] = Seq()
  nodes.foreach { node =>
    val nextTransformer = node match {
      case EstimatorWithData(est, data) => {
        est.fit( PipelineModel(transformers).apply(data) )
      }
      case pipeline: Pipeline => pipeline.fit()
      case transformer: Transformer => transformer
    }
    transformers = transformers :+ nextTransformer
  }

  PipelineModel(transformers)
}
```
It's got more to it, but that's the main idea. Start w/ an empty seq of Transformers. Go through your nodes one by one. If you see a transformer, add it to your transformers as-is. If you see a pipeline, fit it then add the output transformer. If you see an estimator w/ data, first apply your current seq of transformers to the data, then fit the estimator on the transformed data. Finally wrap your Seq of transformers in a PipelineModel (a transformer made up of a chain of transformers)

## What is a Node?
A ```Node``` is anything that can meaningfully be chained as part of a Pipeline (basically anything that ```Pipeline.fit()``` knows what to do with)
Transformers are nodes
A pipeline is a node
The scatter operator is a node
Estimators on their own are NOT nodes, but Estimators w/ data attached are.
LabeledEstimators w/ data attached are also nodes.

Nodes have a `rewrite` method (output a seq of nodes that don't change the fit output when appended onto an existing pipeline as opposed to just outputting the current node, but make the DAG flatter/more interpretable and in the case of scatter nodes avoids repeating work)

Nodes also have a `canSafelyPrependExtraNodes` method that returns true if it is safe for the Pipeline that directly encloses this node to have new nodes prepended. E.g. false for estimators w/ data because the attached data type must match the input type of the enclosing Pipeline, but true for Transformers. For scatter nodes it depends on what they internally contain. 

## What I want feedback on
I'm not entirely happy with how leaky the abstractions are in `Pipeline.fit()` (creating new DAG operators or new types of Nodes requires updating Pipeline.fit() no matter what). Without a better idea of how @etrain wants to do checkpointing stuff I can't quite be sure how to proceed.

I'm also not sure what the right balance is between calling `node.rewrite` everywhere versus having constructors call rewrite but not being able to take advantage of case classes in some cases. E.g. `Pipeline` has a constructor that forces a rewrite which means it can't be a case class, so I have to implement unapply, toString, etc. methods but on the other hand I don't have to put rewrite everywhere to make sure it ends up called. (technically we could get away with only ever calling rewrite right before a fit, but the DAG wouldn't be as clean & predictable when trying to drill down programmatically if we don't call rewrite when constructing or chaining things) 

Some of the method names are kind of ugly (e.g. canSafelyPrependExtraNodes), and if I end up keeping them, suggestions for alternative names are welcome.

(misc note: most of the changed files are just changing imports to be Workflow nodes instead of previous nodes)


etrain:
@tomerk Thanks for this - this is starting to look good.

A couple of points.
1. As I mentioned in our conversation yesterday, I guess I have two main worries - first, I am worried that the data structures we're using internally aren't sufficient to capture the types of DAGs that we want to. The Sample then Estimator pattern which feeds back into the "main" pipeline can't currently be written in a way that the DAG captures that structure - instead we're talking about resorting to some hacks to do this inside the node, which makes it harder for us to optimize. I suspect this is because we haven't fully defined the semantics of the connections between Data, Estimators, and the Transformers that they produce, so maybe that needs to be fixed before we can make real progress here. Basically I'm saying let's get away from assuming these things are purely linear - there are going to be branches and merges along the way, and we should be able to capture that. The second worry is that things are getting too complicated. The presence of a function called "canSafelyPrependExtraNodes", for example, has me worried that something has gotten too tricky.
2. I don't know that changing pipeline.fit() every time you add a new node type is that bad - what *is* bad is if we think we're going to be adding *lots* of new node types that would make this unmaintainable. Basically I think there are probably five or six common patterns we should be supporting at this level, and that we should be able to compose more complicated pipelines out of these patterns.
3. The fact that *pipeline rewriting* is so entwined with the definition of a pipeline also has me a little worried. Essentially I think that pipelines should execute correctly (if naively) without any rewriting, and that the thing that does the rewriting should be looking at the DAG for optimizations it can make to speed some things up - preferably with centralized (if recursive) logic.
4. The "we can't make these case classes" thing is not a big deal because that part is just syntax - and while I like nice, concise syntax, that side of things is fixable. What's more troubling is the idea that the optimization  

Finally, let's make sure we mark all this internal DAG ordering stuff as protected - user code shouldn't be touching it, and it's easier to take away these APIs up front than remove them later.

etrain:
Also - you're much closer to these issues than I am - so if any of what I wrote above doesn't make sense, or if you think I'm missing something, let's have a conversation about it!

tomerk:
Future work happening on my branch dag-construction-v2

PR #143, opened 2015-06-09
Fixing bug in ZCA Whitener where it was using a (large) hardcoded eps…

etrain:
…ilon, and fixing issue in unit test that was hiding this.

CC @tomerk

### DIFF
--- a/src/test/scala/nodes/learning/ZCAWhiteningSuite.scala
+++ b/src/test/scala/nodes/learning/ZCAWhiteningSuite.scala
@@ -8,18 +8,30 @@ import pipelines._
 
 class ZCAWhiteningSuite extends FunSuite with LocalSparkContext with Logging {
 
-  test("whitening") {
-    val eps = 0.1
-    val nrows = 10000
-    val ndim = 10
+  val nrows = 10000
+  val ndim = 10
 
-    val x = DenseMatrix.rand[Double](nrows, ndim, Gaussian(0.0, 1.0))
-    val whitener = new ZCAWhitenerEstimator().fitSingle(x)
+  val x = DenseMatrix.rand[Double](nrows, ndim, Gaussian(0.0, 1.0))
+
+  def fitAndCompare(x: DenseMatrix[Double], eps: Double, thresh: Double): Boolean = {
+    val whitener = new ZCAWhitenerEstimator(eps).fitSingle(x)
 
     val wx = whitener(x)
 
-    //Checks max(max(abs(cov(whiten(x))) - eye(10)) < 2*eps
-    assert(max(abs(cov(convert(wx, Double))) - DenseMatrix.eye[Double](ndim)) < 2*eps,
+    //Checks max(max(abs(cov(whiten(x))) - eye(10)) < sqrt(eps)
+    max(abs(cov(convert(wx, Double))) - DenseMatrix.eye[Double](ndim)) < thresh
+  }
+
+  test("whitening with small epsilon") {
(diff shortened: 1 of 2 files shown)


src/test/scala/nodes/learning/ZCAWhiteningSuite.scala
```diff
@@ -14,7 +14,7 @@ class ZCAWhiteningSuite extends FunSuite with LocalSparkContext with Logging {
     val ndim = 10
 
     val x = DenseMatrix.rand[Double](nrows, ndim, Gaussian(0.0, 1.0))
-    val whitener = new ZCAWhitenerEstimator(eps).fitSingle(x)
```
shivaram: So now `eps` defined above is not getting passed in -- should we just remove the eps definition above ?

Also we could add another test with eps=1e-3 and test whether the code is using it correctly ?

etrain: Well. We use eps in the test itself - but I see your point on varying
parameters. Will cook something up.


On Jun 8, 2015, at 5:40 PM, Shivaram Venkataraman <notifications@github.com>
wrote:

In src/test/scala/nodes/learning/ZCAWhiteningSuite.scala
<https://github.com/amplab/keystone/pull/143#discussion_r31974391>:

> @@ -14,7 +14,7 @@ class ZCAWhiteningSuite extends FunSuite with LocalSparkContext with Logging {
>      val ndim = 10
>
>      val x = DenseMatrix.rand[Double](nrows, ndim, Gaussian(0.0, 1.0))
> -    val whitener = new ZCAWhitenerEstimator(eps).fitSingle(x)

So now eps defined above is not getting passed in -- should we just remove
the eps definition above ?

Also we could add another test with eps=1e-3 and test whether the code is
using it correctly ?

—
Reply to this email directly or view it on GitHub
<https://github.com/amplab/keystone/pull/143/files#r31974391>.

etrain:
I've updated this to make the tests a little more comprehensive - testing at a couple levels of epsilon and making sure that it *doesn't* work when it shouldn't.

shivaram:
Very cool  ! LGTM. Merging this

src/test/scala/nodes/learning/ZCAWhiteningSuite.scala
```diff
@@ -8,18 +8,30 @@ import pipelines._
...
+
+  def fitAndCompare(x: DenseMatrix[Double], eps: Double, thresh: Double): Boolean = {
     val whitener = new ZCAWhitenerEstimator(eps).fitSingle(x)
 
     val wx = whitener(x)
 
-    //Checks max(max(abs(cov(whiten(x))) - eye(10)) < 2*eps
-    assert(max(abs(cov(convert(wx, Double))) - DenseMatrix.eye[Double](ndim)) < 2*eps,
+    //Checks max(max(abs(cov(whiten(x))) - eye(10)) < sqrt(eps)
+    max(abs(cov(convert(wx, Double))) - DenseMatrix.eye[Double](ndim)) < thresh
```
tomerk: So I probably should have commented this earlier, but: the abs should be surrounding the subtraction, not the covariance. Otherwise if covariance is e.g. 0.1 along the diagonals this would still pass even though it shouldn't.

shivaram: Oh good catch - Could you send a PR to fix this ?

tomerk: Yeah, will do tomorrow.

PR #144, opened 2015-06-09
Fixing bug in unit test -> abs(diff(x,y)) as opposed to diff(abs(x),y)

etrain:
Thanks @tomerk for pointing this out.


tomerk:
LGTM

PR #145, opened 2015-06-11
Made what matrix utils I trivially could be generic

tomerk:
@shivaram


shivaram:
LGTM. Thanks @tomerk ! This should make some things a little better

Merging this

Issue #121, opened 2015-05-18
Review use of FunctionNode


etrain:
In several pipelines, we use `FunctionNode` to handle cases where, for example, an `Estimator[A,B]` doesn't return a `Transformer[A,B]`, but instead returns a `Transformer[C,D]`, or where there is no good meaning for a single-item transformation. 

Currently, FunctionNode feels like a "catch-all" because the Transformer/Estimator APIs don't sufficiently cover some of the data transformation operations we need to support.

One example of this is `NGramsCounts` which takes a `Seq[Seq[T]]` and returns a model of type `NGrams[T] => Int`.

Other examples include `Windower` and `Sampler` which are used in the RandomPatchCifar pipeline. These nodes are different in that they do not operate on single items and are thus not transformers, but act as something like an `Aggregator` if we were going to draw a database analogy.

PR #146, opened 2015-06-14
Dag construction v2

tomerk:
@etrain @shivaram Not yet ready for merging (and plenty of cleanup to be done), but feel free to start taking a look


tomerk:
@etrain @shivaram This is now ready for reviewing & merging

src/main/scala/nodes/images/LabeledImageExtractors.scala
```diff
@@ -2,6 +2,7 @@ package nodes.images
 
 import pipelines._
```
etrain: what are we getting from pipelines anymore?

src/main/scala/nodes/stats/PaddedFFT.scala
```diff
@@ -10,7 +10,7 @@ import pipelines.Transformer
  *
  * Goes from vectors of size n to vectors of size nextPositivePowerOfTwo(n)/2
  */
-object PaddedFFT extends Transformer[DenseVector[Double], DenseVector[Double]] {
+case class PaddedFFT() extends Transformer[DenseVector[Double], DenseVector[Double]] {
```
etrain: What's the reason to prefer case class here? Just consistency?

src/main/scala/nodes/stats/Sampling.scala
```diff
@@ -2,7 +2,7 @@ package nodes.stats
 
 import breeze.linalg.{DenseVector, DenseMatrix}
 import org.apache.spark.rdd.RDD
-import pipelines.{FunctionNode, Transformer}
+import pipelines.FunctionNode
```
etrain: Do we still need FunctionNode?

src/main/scala/pipelines/images/cifar/LinearPixels.scala
```diff
@@ -22,8 +22,8 @@ object LinearPixels extends Logging {
     val trainData = CifarLoader(sc, config.trainLocation).cache()
 
     // A featurizer maps input images into vectors. For this pipeline, we'll also convert the image to grayscale.
-    val featurizer = GrayScaler then ImageVectorizer
-    val labelExtractor = LabelExtractor then ClassLabelIndicatorsFromIntLabels(numClasses) then new Cacher[DenseVector[Double]]
+    val featurizer = GrayScaler andThen ImageVectorizer
+    val labelExtractor = LabelExtractor andThen ClassLabelIndicatorsFromIntLabels(numClasses) andThen new Cacher[DenseVector[Double]]
```
etrain: probably want to spread this over more than one line.

src/main/scala/pipelines/images/cifar/RandomCifar.scala
```diff
@@ -33,17 +33,18 @@ object RandomCifar extends Serializable with Logging {
...
-        .then(SymmetricRectifier(alpha=conf.alpha))
-        .then(new Pooler(conf.poolStride, conf.poolSize, identity, _.sum))
-        .then(ImageVectorizer)
-        .then(new Cacher[DenseVector[Double]])
-        .thenEstimator(new StandardScaler).fit(trainImages)
-        .then(new Cacher[DenseVector[Double]])
-
-    val labelExtractor = LabelExtractor then
-      ClassLabelIndicatorsFromIntLabels(numClasses) then
+        .andThen(SymmetricRectifier(alpha=conf.alpha))
```
etrain: can we switch to "andThen" at the end of the line - I greatly prefer this style.

src/main/scala/pipelines/text/NewsgroupsPipeline.scala
```diff
@@ -21,15 +21,17 @@ object NewsgroupsPipeline extends Logging {
...
-    val predictor = Trim.then(LowerCase())
-        .then(Tokenizer())
-        .then(new NGramsFeaturizer[String](1 to conf.nGrams))
-        .then(TermFrequency(x => 1))
-        .thenEstimator(CommonSparseFeatures(conf.commonFeatures))
-        .fit(trainData.data)
-        .thenLabelEstimator(NaiveBayesEstimator(numClasses))
-        .fit(trainData.data, trainData.labels)
-        .then(MaxClassifier)
+    val predictorPipeline = Trim andThen LowerCase() andThen
```
etrain: much prettier!

src/main/scala/workflow/DelegatingTransformer.scala
```diff
@@ -0,0 +1,9 @@
+package workflow
+
+import org.apache.spark.rdd.RDD
+
+class DelegatingTransformer[T](override val label: String) extends TransformerNode[T] {
```
etrain: whoa, this is scary sounding! What does it do? Can we have some comments?

Should this be package protected?

src/main/scala/workflow/Estimator.scala
```diff
@@ -1,21 +1,34 @@
...
-   * An estimator has a `fit` method which emits a pipeline node.
+   * An estimator has a `fit` method which emits a [[Transformer]].
    * @param data Input data.
-   * @return A PipelineNode (Transformer) which can be called on new data.
+   * @return A [[Transformer]] which can be called on new data.
    */
-  def fit(data: RDD[A]): Transformer[A, B]
+  protected def fit(data: RDD[A]): Transformer[A, B]
+
+  private[workflow] final def fit(dependencies: Seq[RDD[_]]): TransformerNode[_] = fit(dependencies.head.asInstanceOf[RDD[A]])
```
etrain: I'll be a little forward here and say I'm worried that throwing away this type information and making assumptions about ordering of the Seq may come back to bite us - but for now I think this is fine.

src/main/scala/workflow/Estimator.scala
```diff
@@ -1,21 +1,34 @@
...
+
+  def withData(data: RDD[A]): Pipeline[A, B] = {
```
etrain: So this is the thing that constructs a pipeline from nothing in case you ant just an estimator pipeline? Can we get a comment to that effect?

src/main/scala/workflow/Gather.scala
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.spark.rdd.RDD
+
+import scala.reflect.ClassTag
+
+private[workflow] class GatherTransformer[T] extends TransformerNode[Seq[T]] {
+  def transform(dataDependencies: Seq[_], fitDependencies: Seq[TransformerNode[_]]): Seq[T] = dataDependencies.map(_.asInstanceOf[T])
+
+  def transformRDD(dataDependencies: Seq[RDD[_]], fitDependencies: Seq[TransformerNode[_]]): RDD[Seq[T]] = {
+    dataDependencies.map(_.asInstanceOf[RDD[T]].map(t => Seq(t))).reduceLeft((x, y) => {
+      x.zip(y).map(z => z._1 ++ z._2)
```
etrain: Is this the most efficient way to combine these Seq's? It seems like you're going to launch a job per RDD?

src/main/scala/workflow/Gather.scala
```diff
@@ -0,0 +1,39 @@
...
+    })
+  }
+}
+
+object Gather {
+  def apply[A, B : ClassTag](branches: Seq[Pipeline[A, B]]): Pipeline[A, Seq[B]] = {
+    val branchesWithNodeOffsets = branches.scanLeft(0)(_ + _.nodes.size).zip(branches)
```
etrain: so to summarize you're assigning a new id for each node in each branch?

src/main/scala/workflow/Node.scala
```diff
@@ -0,0 +1,24 @@
+package workflow
+
+import org.apache.spark.rdd.RDD
+
+sealed trait Node  {
+  def label: String = {
```
etrain: If this is meant to be used for DAG drawing - we have to be careful with parameterization. For example, an NGramsTransformer(2) is different from an NGramsTransformer(1) - and we might have multiple linear mappers in a pipeline, etc. (these are parameterized by their input). Not sure the right way to handle this, though..

src/main/scala/workflow/Optimizer.scala
```diff
@@ -0,0 +1,35 @@
...
+ * Merges equivalent node indices in the DAG.
+ * Nodes are considered equivalent if:
+ * - The nodes at the indices themselves are equal
+ * - They point to the same data dependencies
+ * - They point to the same fit dependencies
+ */
+object EquivalentNodeMerger extends Rule {
+  def apply[A, B](plan: Pipeline[A, B]): Pipeline[A, B] = {
+    val fullNodes = plan.nodes.zip(plan.dataDeps.zip(plan.fitDeps)).zipWithIndex
+    val mergableNodes = fullNodes.groupBy(_._1).mapValues(_.map(_._2)).toSeq
```
etrain: One thing this relies on is that the `Node` themselves have to be `.equal()` to one another for this to work. Probably best to be explicit about this in the comments.

src/main/scala/workflow/Optimizer.scala
```diff
@@ -0,0 +1,35 @@
...
+
+    if (mergableNodes.size == plan.nodes.size) {
+      // no nodes are mergable
+      plan
+    } else {
+      val oldToNewNodeMapping = mergableNodes.zipWithIndex.map(x => x._1._2.map(y => (y, x._2))).flatMap(identity).toMap + (Pipeline.SOURCE -> Pipeline.SOURCE)
```
etrain: break it up on multiple lines, please!

src/main/scala/workflow/Pipeline.scala
```diff
@@ -0,0 +1,197 @@
...
+    this.eq(pipeline) || (
+        (nodes == pipeline.nodes) &&
+        (dataDeps == pipeline.dataDeps) &&
+        (fitDeps == pipeline.fitDeps) &&
+        (sink == pipeline.sink))
+  }
+}
+
+object Pipeline {
+  val SOURCE: Int = -1
```
etrain: This is a little weird, and I kind of feel like we're over-fitting toward the single-input case, but I'm having a tough time coming up with cases where we won't have a single input at estimated-pipeline execution time.

src/main/scala/workflow/Pipeline.scala
```diff
@@ -0,0 +1,197 @@
...
+
+  def apply[T](): Pipeline[T, T] = new ConcretePipeline(Seq(), Seq(), Seq(), SOURCE)
+  private[workflow] def apply[A, B](nodes: Seq[Node], dataDeps: Seq[Seq[Int]], fitDeps: Seq[Seq[Int]], sink: Int): Pipeline[A, B] = new ConcretePipeline(nodes, dataDeps, fitDeps, sink)
+}
+
+private[workflow] class ConcretePipeline[A, B](
```
etrain: Can you say a little about how a `ConcretePipeline` differs from a non Concrete one in the comments?

src/main/scala/workflow/RuleExecutor.scala
```diff
@@ -0,0 +1,85 @@
+package workflow
+
+import pipelines.Logging
+
+abstract class RuleExecutor extends Logging {
```
etrain: The RuleExecutor - we're lifting from Catalyst, right? We can't use the Catalyst one directly because it's marked private or some other reason? (I really hate copy-pasting code!)

src/main/scala/workflow/Rule.scala
```diff
@@ -0,0 +1,14 @@
+package workflow
+
+/**
+ * Represents a pipeline rule
```
etrain: And a rule is what, exactly?

etrain:
This is looking great @tomerk - thanks for the awesome work, and for going through several iterations of this - I think we're honing in on something great!

Most of my comments are stylistic at this point, but I've also been using this branch to work on pipeline-rule rewrites and found the datastructures/APIs pretty usable. 

After a round of style cleanup, given the size and importance of this patch, I'm going to ask @shivaram to  review it as well!

src/main/scala/workflow/Pipeline.scala
```diff
@@ -0,0 +1,197 @@
...
+          transformer.transform(nodeDataDeps, nodeFitDeps)
+        case _: DataNode =>
+          throw new RuntimeException("Pipeline DAG error: came across an RDD data dependency when trying to do a single item apply")
+        case _: EstimatorNode =>
+          throw new RuntimeException("Pipeline DAG error: Cannot have a data dependency on an Estimator")
+      }
+    }
+  }
+
+  private def rddDataEval(node: Int, in: RDD[A]): RDD[_] = {
```
etrain: I'm planning to use this (and possibly its non-RDD brother) in my caching stuff. Any good reason not to mark it private[workflow]?

src/main/scala/nodes/images/LabeledImageExtractors.scala
tomerk: ah, I'm guessing nothing, I'll remove this.

src/main/scala/nodes/stats/PaddedFFT.scala
tomerk: Yeah, just consistency.

src/main/scala/nodes/stats/Sampling.scala
tomerk: At the moment it's still implemented as a FunctionNode, but we can change it to a Transformer and not need FunctionNode.

etrain:
Also - would be great to get some unit tests on this? With all the Type `_` floating around, especially!

src/main/scala/pipelines/images/cifar/LinearPixels.scala
tomerk: Will do!

src/main/scala/pipelines/images/cifar/RandomCifar.scala
tomerk: Will do!

src/main/scala/workflow/DelegatingTransformer.scala
tomerk: This should probably be package protected. This basically just takes in a single fit dependency, and applies that fit dependency on the input data. It's what we currently chain estimator outputs to. I'll add some comments.

src/main/scala/workflow/Estimator.scala
tomerk: I'm less concerned about the ordering of seq mattering (as I'm just thinking of it as representing function parameters in an untyped way, where order still matters). We really do have to be careful because we're throwing all the types away, and I could be convinced of Transformer1-** and Estimator1-** being the right thing to do instead if we can find an alternative to how the Gather operator is currently done.

src/main/scala/workflow/Estimator.scala
tomerk: Sure, I'll add a comment.

src/main/scala/workflow/Gather.scala
tomerk: I'm not sure. It looks like spark has zipPartitions up to 4 rdds at once, so we could do a little bit better than one job per RDD. With the way the DAG execution graph is being represented here though we can't really pick the alternate execution plans of the preceding nodes (meaning we apply each of the previous paths' single apply and do it all on a single RDD).

src/main/scala/workflow/Gather.scala
tomerk: Yes. Shifted by however many nodes are in the preceding branches

src/main/scala/workflow/Node.scala
tomerk: I didn't want to just call toString because that would get unwieldy w/ case classes parameterized by DenseMatrices/DenseVectors, but we could do some reflection-esque stuff to extract all 'small' parameters.

src/main/scala/workflow/Optimizer.scala
tomerk: Yeah I'll make that clearer in this line: ` - The nodes at the indices themselves are equal`

src/main/scala/workflow/Optimizer.scala
tomerk: Will do!

src/main/scala/workflow/Pipeline.scala
tomerk: This very well may have been me overfitting DSL stuff into the core DAG representation, but like you mention I can't think of any cases where the final pipeline doesn't have a single input.

src/main/scala/workflow/Pipeline.scala
tomerk: Will do!

src/main/scala/workflow/RuleExecutor.scala
tomerk: I'm not using the Catalyst one directly here because it is parameterized on TreeNode[_], which pipelines don't really make sense as. It may be possible to force all Pipelines (including Transformers) to extend TreeNode[Pipeline] so we can use the built-in catalyst one (it's unclear, would need to dig into it a little bit).

src/main/scala/workflow/Rule.scala
tomerk: Transforms a Pipeline to a functionally-equivalent, different Pipeline. will extend the comments.

src/main/scala/workflow/Pipeline.scala
tomerk: mmm not really. I can mark it `final private[workflow]`

tomerk:
@etrain I'll go about making these stylistic changes, and I'll see what I can do about unit tests (let me know if you have any ideas).
Also, if you haven't done so yet feel free to look at my proposal for the next version of this stuff that brings it more in line with standard practices for optimizing compilers: https://gist.github.com/tomerk/2f0021d005ac3e18d535
So long as this is usable enough for you though (and we're careful about avoiding the tricky determinism bugs we previously discussed) there's no need to start on that yet.

src/main/scala/nodes/stats/Sampling.scala
etrain: Any reason not to do this?

etrain:
Re: Unit tests - it probably makes sense to construct very simple pipelines with a few different shapes, and make sure that 1) they don't NPE on construction or execution, and 2) that rule transformations "work" the way you think they should.

src/main/scala/nodes/stats/Sampling.scala
tomerk: I was just putting it off to a different PR when we clean up the existing pipelines. If you want me to do it now I can.

tomerk:
@etrain I've now responded to your comments & added unit tests

shivaram:
@etrain let me know once this is good for me to take a look

src/main/scala/nodes/stats/Sampling.scala
shivaram: FWIW We should create a new issue to track this -- our issue tracker could use some love :)

src/main/scala/nodes/util/VectorCombiner.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import breeze.linalg.{DenseMatrix, DenseVector}
+import workflow.Transformer
+
+import scala.reflect.ClassTag
+
+/**
+ * Concats a Seq of DenseVectors into a single DenseVector.
+ */
+case class VectorCombiner[T : ClassTag]()(implicit zero: breeze.storage.Zero[T]) extends Transformer[Seq[DenseVector[T]], DenseVector[T]] {
```
shivaram: over 100 characters ?

tomerk: Ack, 140 chars, will fix.

src/main/scala/nodes/stats/Sampling.scala
tomerk: Already exists, #121 :)

src/main/scala/pipelines/images/mnist/MnistRandomFFT.scala
```diff
@@ -40,48 +35,32 @@ object MnistRandomFFT extends Serializable with Logging {
         .cache())
     val labels = ClassLabelIndicatorsFromIntLabels(numClasses).apply(train.labels)
 
-    val batchFeaturizer = (0 until numFFTBatches).map { batch =>
-      (0 until fftsPerBatch).map { x =>
-        RandomSignNode(mnistImageSize, randomSignSource) then PaddedFFT then LinearRectifier(0.0)
+    val featurizer = Pipeline.gather {
```
shivaram: Nice ! this is sweet

src/main/scala/workflow/DelegatingTransformer.scala
```diff
@@ -0,0 +1,13 @@
...
+import org.apache.spark.rdd.RDD
+
+/**
+ * A transformer used internally to apply the fit output of an Estimator onto data dependencies.
+ * Takes one fit dependency, and directly applies its transform onto the data dependencies.
+ */
+private[workflow] class DelegatingTransformer[T](override val label: String) extends TransformerNode[T] {
+  def transform(dataDependencies: Seq[_], fitDependencies: Seq[TransformerNode[_]]): T = fitDependencies.head.transform(dataDependencies, Seq()).asInstanceOf[T]
+
+  def transformRDD(dataDependencies: Seq[RDD[_]], fitDependencies: Seq[TransformerNode[_]]): RDD[T] = fitDependencies.head.transformRDD(dataDependencies, Seq()).asInstanceOf[RDD[T]]
```
shivaram: again line length limits.

src/main/scala/workflow/DelegatingTransformer.scala
```diff
@@ -0,0 +1,13 @@
+package workflow
+
+import org.apache.spark.rdd.RDD
+
+/**
+ * A transformer used internally to apply the fit output of an Estimator onto data dependencies.
+ * Takes one fit dependency, and directly applies its transform onto the data dependencies.
+ */
+private[workflow] class DelegatingTransformer[T](override val label: String) extends TransformerNode[T] {
+  def transform(dataDependencies: Seq[_], fitDependencies: Seq[TransformerNode[_]]): T = fitDependencies.head.transform(dataDependencies, Seq()).asInstanceOf[T]
```
shivaram: Could we document above that we use the first fitDependency as that seems kind of important ? Also some reasoning as to why that is okay would be good to have

src/main/scala/workflow/GatherTransformer.scala
```diff
@@ -0,0 +1,15 @@
...
+import org.apache.spark.rdd.RDD
+
+import scala.reflect.ClassTag
+
+private[workflow] class GatherTransformer[T] extends TransformerNode[Seq[T]] {
+  def transform(dataDependencies: Seq[_], fitDependencies: Seq[TransformerNode[_]]): Seq[T] = dataDependencies.map(_.asInstanceOf[T])
+
+  def transformRDD(dataDependencies: Seq[RDD[_]], fitDependencies: Seq[TransformerNode[_]]): RDD[Seq[T]] = {
+    dataDependencies.map(_.asInstanceOf[RDD[T]].map(t => Seq(t))).reduceLeft((x, y) => {
+      x.zip(y).map(z => z._1 ++ z._2)
```
shivaram: This is probably not a big deal, but from what I can see this sets up a lineage of length `n` if we have `n` dataDependencies. I think we can make it `log n` if we used a tree-like structure, but it might not matter for our use cases

etrain: @tomerk and I discussed this yesterday. We used to have a Zip `FunctionNode` that did this, and agreed that it might be possible to make the DAG smaller. Should we add an issue? In general I think we're looking at n < 100 for most of our uses.

src/main/scala/workflow/Pipeline.scala
```diff
@@ -0,0 +1,242 @@
+package workflow
+
+import org.apache.spark.rdd.RDD
+import pipelines.Logging
+
+import scala.collection.mutable
+import scala.reflect.ClassTag
+
+trait Pipeline[A, B] {
+  private[workflow] val nodes: Seq[Node]
```
shivaram: These member variables need comments -- what is a fitDep, what is a dataDep etc. In fact an over-arching comment like `each pipeline consists of a number of nodes and a special sink node ...`  might be useful

src/main/scala/workflow/Pipeline.scala
```diff
@@ -0,0 +1,242 @@
...
+
+    - data deps may not point at estimators
+    - fit deps may only point to estimators
+    */
+  private[workflow] def validate(): Unit = {
+    require(nodes.size == dataDeps.size && nodes.size == fitDeps.size, "nodes.size must equal dataDeps.size and fitDeps.size")
+    require(sink == Pipeline.SOURCE || (sink >= 0 && sink < nodes.size), "Sink must point at a valid node")
+
+    val nodeTuples = nodes.zip(dataDeps).zip(fitDeps).map(x => (x._1._1, x._1._2, x._2))
+    require(nodeTuples.forall(x => if (x._1.isInstanceOf[DataNode] && (x._2.nonEmpty || x._3.nonEmpty)) false else true), "DataNodes may not have dependencies")
```
shivaram: lots of long lines here :)

src/main/scala/workflow/Pipeline.scala
```diff
@@ -0,0 +1,242 @@
...
+      fitDeps.map(_.map(x => if (x == Pipeline.SOURCE) Pipeline.SOURCE else x + offset))
+    }.reduceLeft(_ ++ _) :+  Seq()
+
+    val newSink = newNodes.size - 1
+    Pipeline(newNodes, newDataDeps, newFitDeps, newSink)
+  }
+
+}
+
+/**
```
shivaram: Can we pull this out into a new file ? It looks like a big enough class to do that

src/main/scala/workflow/Pipeline.scala
tomerk: Yeah, I'm also not sure about leaving this method in general. It could be useful, but I couldn't make it automatically run at construction time due to super constructor ordering stuff. Maybe I should move it to `ConcretePipeline`?

src/main/scala/workflow/Optimizer.scala
```diff
@@ -0,0 +1,37 @@
+package workflow
+
+/**
+ * Optimizes a Pipeline DAG
+ */
+object Optimizer extends RuleExecutor {
+  protected val batches: Seq[Batch] = Batch("DAG Optimization", FixedPoint(100), EquivalentNodeMerger) :: Nil
```
shivaram: Just curious -- for the pipelines we have today, how useful is merging equivalent nodes ?

etrain: @tomerk can correct me - but this is basically moving pipelines from a tree structure with a lot of redundant branches into a more compact DAG that eliminates much of this redundancy. Execution of the pipeline is basically memo-ized depth-first search from the result in this DAG, so consolidating the representation helps a bunch!

shivaram:
@tomerk this is looking pretty good at a high level and I just had some minor comments. I think just explaining the Pipeline data structures a bit more would make this nice and clear. 

@etrain One more thing we should do is to tag the previous version as 0.1 as this is a pretty big API change.

tomerk:
@shivaram how's this look now?

src/main/scala/pipelines/text/NewsgroupsPipeline.scala
```diff
@@ -21,15 +21,17 @@ object NewsgroupsPipeline extends Logging {
...
+        Tokenizer() andThen
+        NGramsFeaturizer(1 to conf.nGrams) andThen
+        TermFrequency(x => 1) andThen
+        (CommonSparseFeatures(conf.commonFeatures), trainData.data) andThen
+        (NaiveBayesEstimator(numClasses), trainData.data, trainData.labels) andThen
+        MaxClassifier
+
+
+    val predictor = Optimizer.execute(predictorPipeline)
```
shivaram: Do we need to call the optimizer explicitly or is this just done here to get the dot file ?

tomerk: Whenever you want to use the optimizer it has to be called explicitly

src/main/scala/workflow/Estimator.scala
```diff
@@ -1,21 +1,40 @@
...
+   * Equivalent to `Pipeline() andThen (estimator, data)`
+   *
+   * @param data The training data
    */
-  def fit(data: RDD[A]): Transformer[A, B]
+  def withData(data: RDD[A]): Pipeline[A, B] = {
+    val nodes: Seq[Node] = Seq(DataNode(data), this, new DelegatingTransformer[B](this.label + ".fit"))
+    val dataDeps = Seq(Seq(), Seq(0), Seq(Pipeline.SOURCE))
+    val fitDeps = Seq(Seq(), Seq(), Seq(1))
+    val sink = 2
```
shivaram: Can we make this also a constant in the Pipeline object ? (similar to Pipeline.SOURCE)

src/main/scala/workflow/LabelEstimator.scala
```diff
@@ -1,23 +1,43 @@
...
+   * Equivalent to `Pipeline() andThen (estimator, data, labels)`
+   *
+   * @param data The training data
+   * @param labels The training labels
+   */
+  def withData(data: RDD[A], labels: RDD[L]): Pipeline[A, B] = {
+    val nodes: Seq[Node] = Seq(DataNode(data), DataNode(labels), this, new DelegatingTransformer[B](this.label + ".fit"))
+    val dataDeps = Seq(Seq(), Seq(), Seq(0, 1), Seq(Pipeline.SOURCE))
+    val fitDeps = Seq(Seq(), Seq(), Seq(), Seq(2))
+    val sink = 3
```
shivaram: Can we make this one a constant as well ? (Giving it a descriptive name would also be good)

src/main/scala/workflow/Transformer.scala
```diff
@@ -0,0 +1,53 @@
...
+  def apply(in: RDD[A]): RDD[B] = in.map(apply)
+
+  /**
+   * Apply this Transformer to a single input item
+   * @param in  The input item to pass into this transformer
+   * @return  The output value
+   */
+  def apply(in: A): B
+
+  private[workflow] final def transform(dataDependencies: Seq[_], fitDependencies: Seq[TransformerNode[_]]): B = apply(dataDependencies.head.asInstanceOf[A])
```
shivaram: long lines -- is there a way you can ask intelli-j to fix all of these for you ?

src/test/scala/workflow/DelegatingTransformerSuite.scala
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  test("rdd apply") {
+    sc = new SparkContext("local", "test")
+
+    val hashTransformer = Transformer[String, Int](_.hashCode)
+    val delegatingTransformer = new DelegatingTransformer[Int]("label")
+
+    val strings = Seq("A31DFSsafds*be31", "lj32fsd", "woadsf8923")
+    val transformedStrings = delegatingTransformer.transformRDD(Seq(sc.parallelize(strings)), Seq(hashTransformer)).collect()
```
shivaram: long line

src/test/scala/workflow/PipelineSuite.scala
```diff
@@ -0,0 +1,101 @@
...
+        Transformer(x => x + first)
+      }
+    }, sc.parallelize(Seq(32, 94, 12)), sc.parallelize(Seq("10", "7", "14")))
+
+    val pipeline = Pipeline.gather {
+      firstPipeline :: secondPipeline :: thirdPipeline :: Nil
+    }
+
+    val single = 7
+    assert(pipeline(single) === Seq(firstPipeline.apply(single), secondPipeline.apply(single), thirdPipeline.apply(single)))
```
shivaram: long line

shivaram:
Just some minor style comments. A couple of obligatory questions

1. Do the tests pass ?
2. Have you run some of the pipelines (say Mnist / Cifar) to see if nothing becomes slower etc ?

src/main/scala/pipelines/text/NewsgroupsPipeline.scala
shivaram: Does this mean we should add this to other pipelines as well now ? Or is it okay to not call the optimizer ? Also longer term we shouldn't expect the user to call this I guess, so we need to find a place to do this automatically

src/main/scala/workflow/Estimator.scala
tomerk: Mmm I don't think this makes sense as a constant in the Pipeline object. It's only two here as an artifact of the nodes Seq being constructed above having the DelegatingTransformer as the last item, it doesn't hold any special meaning beyond this specific graph construction. I could replace this with `nodes.size + 1` if you prefer, or add some comments explaining it?

src/main/scala/workflow/LabelEstimator.scala
tomerk: Same as above

src/main/scala/workflow/Estimator.scala
shivaram: Yeah `nodes.size + 1` will be much better than `2`

src/main/scala/workflow/Transformer.scala
tomerk: Don't think so :(

src/main/scala/pipelines/text/NewsgroupsPipeline.scala
tomerk: Mmm... at the moment it will only make a difference in pipelines with > 1 estimator. We should add it to pipelines once we update them to construct the full DAGs before doing anything, but most of them currently don't.

As for when to call this automatically... still tbd, possibly when interpreting or compiling a pipeline, and @etrain may have some ideas.

src/test/scala/workflow/DelegatingTransformerSuite.scala
tomerk: will fix

src/test/scala/workflow/PipelineSuite.scala
tomerk: will fix

tomerk:
Yup, the tests pass!
As for running the pipelines: I have run the newsgroups & Mnist pipelines to confirm they still work correctly. I haven't been measuring performance or running on large datasets, but nothing became obviously slower.

tomerk:
Okay @shivaram, what about now?

shivaram:
LGTM. Thanks @tomerk for working through this PR !

PR #147, opened 2015-06-25
Made the sparse feature estimators deterministic

tomerk:
(to prevent bugs when repeated in the Pipeline DAG)

@etrain @shivaram 
I think these are the only estimators/transformers that are currently non-deterministic (other than the K-means++ in the imagenet research branch)
If I'm wrong about that we should fix whatever other nodes aren't too.

### DIFF
--- a/src/main/scala/nodes/util/AllSparseFeatures.scala
+++ b/src/main/scala/nodes/util/AllSparseFeatures.scala
@@ -8,12 +8,20 @@ import scala.reflect.ClassTag
 
 /**
  * An Estimator that chooses all sparse features observed when training,
- * and produces a transformer which builds a sparse vector out of them
+ * and produces a transformer which builds a sparse vector out of them.
+ *
+ * Deterministically orders the feature mappings by earliest appearance in the RDD
  */
 case class AllSparseFeatures[T: ClassTag]() extends Estimator[Seq[(T, Double)], SparseVector[Double]] {
   override def fit(data: RDD[Seq[(T, Double)]]): SparseFeatureVectorizer[T] = {
-    val featureSpace = data.flatMap(_.map(_._1)).distinct()
-        .zipWithIndex().collect().map(x => (x._1, x._2.toInt)).toMap
+    val featureOccurrences = data.flatMap(_.map(_._1))
+    // zip with unique ids and take the smallest unique id for a given feature to get
+    // a deterministic ordering
+    val featuresWithUniqueId = featureOccurrences.zipWithUniqueId().reduceByKey {
+      (x, y) => Math.min(x, y)
+    }
+    val featureSpace = featuresWithUniqueId.sortBy(_._2).map(_._1)
+        .collect().zipWithIndex.toMap
     new SparseFeatureVectorizer(featureSpace)
   }
 }
--- a/src/main/scala/nodes/util/CommonSparseFeatures.scala
+++ b/src/main/scala/nodes/util/CommonSparseFeatures.scala
@@ -10,17 +10,31 @@ import scala.reflect.ClassTag
  * An Estimator that chooses the most frequently observed sparse features when training,
  * and produces a transformer which builds a sparse vector out of them
  *
+ * Deterministically orders the feature mappings first by decreasing number of appearances,
+ * then by earliest appearance in the RDD
+ *
  * @param numFeatures The number of features to keep
  */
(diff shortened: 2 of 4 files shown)


src/main/scala/nodes/util/AllSparseFeatures.scala
```diff
@@ -8,12 +8,20 @@ import scala.reflect.ClassTag
...
  */
 case class AllSparseFeatures[T: ClassTag]() extends Estimator[Seq[(T, Double)], SparseVector[Double]] {
   override def fit(data: RDD[Seq[(T, Double)]]): SparseFeatureVectorizer[T] = {
-    val featureSpace = data.flatMap(_.map(_._1)).distinct()
-        .zipWithIndex().collect().map(x => (x._1, x._2.toInt)).toMap
+    val featureOccurrences = data.flatMap(_.map(_._1))
+    // zip with unique ids and take the smallest unique id for a given feature to get
+    // a deterministic ordering
+    val featuresWithUniqueId = featureOccurrences.zipWithUniqueId().reduceByKey {
+      (x, y) => Math.min(x, y)
```
etrain: What if we make `T <:< Ordering` (or something) and then just do something like
```
 val featureSpace = data.flatMap(_.map(_._1)).distinct()
        .collect().sorted.zipWithIndex.map(x => (x._1, x._2.toInt)).toMap
```
We pay a local sort, but no shuffle, and I don't know how much this is used in practice.

src/main/scala/nodes/util/CommonSparseFeatures.scala
```diff
@@ -10,17 +10,31 @@ import scala.reflect.ClassTag
...
  * and produces a transformer which builds a sparse vector out of them
  *
+ * Deterministically orders the feature mappings first by decreasing number of appearances,
+ * then by earliest appearance in the RDD
+ *
  * @param numFeatures The number of features to keep
  */
 case class CommonSparseFeatures[T : ClassTag](numFeatures: Int) extends Estimator[Seq[(T, Double)], SparseVector[Double]] {
   // Ordering that compares (feature, frequency) pairs according to their frequencies
-  val ordering = new Ordering[(T, Int)] {
```
etrain: Similarly, if `T <:< Ordering` then we may just be able to do something like `(x._2, x._1) compare (y._2, y._1)` here and get rid of the `zipWithUniqueId` stuff below.

Are there cases where we expect `T` not to extend `Ordering`?

src/main/scala/nodes/util/AllSparseFeatures.scala
tomerk: I tried that earlier, but not enough things have an implicit ordering for this to work nicely & easily interpretable w/ our type inference (e.g. Seq, Tuple, etc.). I could replace reduceByKey w/ a localReduceByKey and then the sort would also be local, but more would go to the driver.

etrain:
Did you verify that LDA is deterministic as well?

src/main/scala/nodes/util/CommonSparseFeatures.scala
tomerk: zipWithUniqueId luckily doesn't spawn a new job. And yes, newsgroups pipeline has Seq[String] (ngrams) as T, which doesn't have ordering.

tomerk:
So far as I could tell LDA is deterministic (I was checking the test on the Iris dataset). I also made sure that the eigenvector sort by eigenvalue is stable, so that's deterministic too. That doesn't rule 100% out non-deterministic eigen decomposition in Breeze, but it's at least consistent on the Iris dataset.

src/main/scala/nodes/util/CommonSparseFeatures.scala
etrain: Ugh, right. You'd think "lexicographic" would be a sensible way to order Seq's for `Seq[T <:< Ordering]` - we could do something like this: http://stackoverflow.com/questions/3137918/how-do-i-sort-a-collection-of-lists-in-lexicographic-order-in-scala - but I digress..

The main thing I'm worried about is the stability of this solution - e.g. whether .zipWithUniqueId is guaranteed to give the same results on the same RDD forevermore, but the docs are convincing enough, I guess.

etrain:
Alright - sounds good. Let's keep this in mind in case we spot other funky behavior.

src/main/scala/nodes/util/AllSparseFeatures.scala
etrain: Ok - I'm sold.

PR #148, opened 2015-06-26
Made VOC & Imagenet pipelines fully build a DAG

tomerk:
@shivaram 
@etrain 

I haven't actually tested it to make sure the pipelines don't break... but it at least visualizes correctly


tomerk:
@shivaram @etrain Is this good to merge?

shivaram:
@tomerk Given that we now have scripts to run things, can we just run the imagenet pipeline with say 4k features and make sure it take similar time / gives similar error as before ?

tomerk:
@shivaram We've now verified timings & test error match, feel free to merge

PR #153, opened 2015-07-27
Following redirects and failing on 404 with missing jars.

etrain:
This patch fixes issues with corrupt jar files when trying to build a fresh checkout of KeystoneML according to the instructions on the project webpage.

I went with the minimal set of changes to get this working again here.

We could also more closely mirror the sbt scripts of Spark, but that was a more invasive change than I wanted to make to get this bugfix in.

Thanks @ericmjonas and Dimp Bhat for reporting.

cc @shivaram @tomerk

### DIFF
--- a/sbt/sbt
+++ b/sbt/sbt
@@ -30,9 +30,9 @@ if [ ! -f ${JAR} ]; then
   # Download
   printf "Attempting to fetch sbt\n"
   if hash curl 2>/dev/null; then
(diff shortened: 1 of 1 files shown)


tomerk:
@etrain have you tested that the new code works correctly in both cases?

tomerk:
After checking with @etrain , looks good to me.

PR #154, opened 2015-08-12
Make BlockWeightedLeastSquares work with missing data for classes

shivaram:
We were assuming all the classes exist wherever we did `collects`. Also added a new unit test which has data only from 1 class.


etrain:
LGTM

PR #157, opened 2015-09-16
SBT updates for sonatype release

tomerk:
@etrain @shivaram

### DIFF
--- a/bin/run-main.sh
+++ b/bin/run-main.sh
@@ -28,8 +28,8 @@ if [ -z "$1" ]; then
 fi
 
 ASSEMBLY_JAR=""
-if [ -e "$FWDIR"/target/scala-$SCALA_VERSION/keystone-assembly-*.jar ]; then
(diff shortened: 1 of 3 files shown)


etrain:
This looks good to me, but I guess any scripts that rely on the name of the keystone assembly jar will also need to change?

tomerk:
Oh that's true. I've updated the scripts here now, but we probably also need to update the keystone ec2 scripts.

shivaram:
Alright I'm merging this on master branch. @tomerk You can create a branch-0.1 and backport this as we discussed. 

Also I checked the EC2 scripts and they don't actually have the JAR name anywhere so I think they should be fine.

tomerk:
Wait, don't they copy it to all the slaves?



—
Best Wishes,
Tomer Kaftan

shivaram:
They just do `/root/spark-ec2/copy-dir /root/keystone`

PR #163, opened 2015-09-18
Fix the mlmatrix dependency issue in build.sbt

zhaozhang:
PR test passed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -34,11 +34,12 @@ libraryDependencies ++= Seq(
   val sparkVersion =
     scala.util.Properties.envOrElse("SPARK_VERSION", defaultSparkVersion)
   val excludeHadoop = ExclusionRule(organization = "org.apache.hadoop")
+  val excludeSpark = ExclusionRule(organization = "org.apache.spark")
   libraryDependencies ++= Seq(
     "org.apache.spark" % "spark-core_2.10" % sparkVersion excludeAll(excludeHadoop),
     "org.apache.spark" % "spark-mllib_2.10" % sparkVersion excludeAll(excludeHadoop),
     "org.apache.spark" % "spark-sql_2.10" % sparkVersion excludeAll(excludeHadoop),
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -38,7 +37,8 @@ libraryDependencies ++= Seq(
   libraryDependencies ++= Seq(
     "org.apache.spark" % "spark-core_2.10" % sparkVersion excludeAll(excludeHadoop),
     "org.apache.spark" % "spark-mllib_2.10" % sparkVersion excludeAll(excludeHadoop),
-    "org.apache.spark" % "spark-sql_2.10" % sparkVersion excludeAll(excludeHadoop)
+    "org.apache.spark" % "spark-sql_2.10" % sparkVersion excludeAll(excludeHadoop),
+    "edu.berkeley.cs.amplab" % "mlmatrix_2.10" % "0.1" excludeAll(excludeHadoop)
```
shivaram: Now that I think about it its also better to exclude Spark here as we don't want to inherit Spark from ml-matrix. Can you add another exclusion rule above and add that here ? The syntax will be `excludeAll(excludeHadoop, excludeSpark)`

build.sbt
```diff
@@ -38,7 +37,8 @@ libraryDependencies ++= Seq(
...
   )
```
shivaram: Breeze is another one which is a shared dependency between this project and ml-matrix (See https://github.com/amplab/ml-matrix/blob/master/build.sbt) -- It might be fine to not have Breeze though

cc @etrain

build.sbt
zhaozhang: Trying now.

Zhao

build.sbt
tomerk: I think the breeze declared here will just clobber the breeze from ml matrix (either that or it'll choose the newer version)

shivaram: So when things conflict maven and SBT have different resolution rules AFAIK. One of them picks the higher version while the other picks the first thing it sees. Which is which I forget

shivaram:
LGTM. The breeze thing shouldn't be a problem in either way of resolution I think. Thanks @zhaozhang for digging into this

Issue #165, opened 2015-09-24
Move Amazon example pipeline to master


etrain:
We've got a cool pipeline for the Amazon product reviews dataset, let's move this to master!

etrain:
cc @dcrankshaw

tomerk:
Which pipeline do you want me to move over? The one that's the same as the 20newsgroups but with logistic regression instead of naive bayes at the end?

etrain:
The data loader is also useful - but let's do logistic regression for some
variety.

tomerk:
Do you also want the n-gram hashing amazon pipeline?

etrain:
Closed by #166

PR #166, opened 2015-09-24
Moved over the Amazon pipeline

tomerk:
issue #165 
@etrain

### DIFF
--- a/src/main/scala/pipelines/text/AmazonReviewsPipeline.scala
+++ b/src/main/scala/pipelines/text/AmazonReviewsPipeline.scala
@@ -15,8 +15,6 @@ object AmazonReviewsPipeline extends Logging {
   val appName = "AmazonReviewsPipeline"
 
   def run(sc: SparkContext, conf: AmazonReviewsConfig) {
-
-    logInfo("PIPELINE TIMING: Started training the classifier")
     val amazonTrainData = AmazonReviewsDataLoader(sc, conf.trainLocation, conf.threshold).labeledData
     val trainData = LabeledData(amazonTrainData.repartition(conf.numParts).cache())
 
@@ -24,7 +22,6 @@ object AmazonReviewsPipeline extends Logging {
     val labels = trainData.labels
 
     // Build the classifier estimator
-    logInfo("Training classifier")
     val predictorPipeline = Trim andThen
         LowerCase() andThen
         Tokenizer() andThen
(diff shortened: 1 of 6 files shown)


tomerk:
Okay @etrain how's this look now?

src/main/scala/pipelines/text/AmazonReviewsPipeline.scala
```diff
@@ -0,0 +1,90 @@
...
+import pipelines.Logging
+import scopt.OptionParser
+import workflow.Optimizer
+
+object AmazonReviewsPipeline extends Logging {
+  val appName = "AmazonReviewsPipeline"
+
+  def run(sc: SparkContext, conf: AmazonReviewsConfig) {
+
+    logInfo("PIPELINE TIMING: Started training the classifier")
```
etrain: This probably doesn't need to be there (nor do its counterparts at the end.)

src/main/scala/pipelines/text/NewsgroupsPipeline.scala
```diff
@@ -21,7 +21,8 @@ object NewsgroupsPipeline extends Logging {
 
     // Build the classifier estimator
     logInfo("Training classifier")
-    val predictorPipeline = Trim andThen LowerCase() andThen
+    val predictorPipeline = Trim andThen
+        LowerCase() andThen
```
etrain: The `()` need to be here?

tomerk: Yup, see my previous comment.

etrain:
So I think the main thing is going to be moving over a reasonable cache-oblivious Logistic Regression port, but other than that, the extraneous log statements, and the little formatting (possibly non-) issues, I think this is good to go.

src/main/scala/pipelines/text/AmazonReviewsPipeline.scala
tomerk: Aww you don't like my timing logging?

src/main/scala/pipelines/text/NewsgroupsPipeline.scala
etrain: Ah, right. Stupid scala.

tomerk:
Okay, how does this look now? (using copied code instead of reflection)

tomerk:
@etrain

tomerk:
Okay @etrain how about now? I didn't figure out a way to override the numFeatures directly via the constructor params. This does compile because the only things it uses from MLlib are marked DeveloperAPI. I haven't tested this code though, so not 100% positive yet that it runs (and we should have a unit test for the logistic regression too)

etrain:
Well - lets kill two birds with one stone and use the unit test to verify? An assert on the number of rdds in cache would probably be good too in that one.

tomerk:
Okay @etrain we have the unit test to verify that the logistic regression works (ported over from mllib). Not sure how to test the number of rdds in cache though...

tomerk:
Also @shivaram

etrain:
Re: number of RDDs in cache - see SparkContext.getRDDStorageInfo

tomerk:
Hmm so it looks like the "addIntercept" that the tests use also forces a cache to occur, but it's not a mandatory parameter for the node itself.

etrain:
So update the unit test?

tomerk:
Okay, I did what we talked about and removed the addIntercept param.

Issue #167, opened 2015-09-30
Update website/API docs to reflect 0.2


etrain:
Can we update the website with the latest API documentation and make sure that the syntax on the front-page is what's in the current newsgroups pipeline? 

I've got a few other cosmetic improvements that I'll make once this is done.

etrain:
It probably also makes sense to start versioning the "api" directory in documentation like `api/0.2/` and maybe have a symlink which points `api/latest/` to `api/0.2/`. `api/latest` can then be the canonical place to find the documentation for the latest release.

tomerk:
Silly question, but where can I find the code for the website?

shivaram:
@tomerk Its in the gh-pages branch https://github.com/amplab/keystone/tree/gh-pages

tomerk:
This was merged already so I'll close.

PR #168, opened 2015-10-01
Api 2.0 update

tomerk:
@etrain @shivaram 
issue #167


etrain:
The API link in _layouts/default.html needs to change, but otherwise this looks good. Thanks @tomerk!

tomerk:
Okay @etrain how's it look now?

Issue #170, opened 2015-10-05
Optimization should happen automatically


etrain:
Users shouldn't have to worry about optimizing their pipelines. The system should handle this automatically. It would be good to have `Optimizer.execute(pipeline)` happen automatically as a first step of `pipeline.apply(inputRdd)` - or something. 

Andy major reasons why this can't work?

tomerk:
We'll need to prevent an infinite recursion (but that shouldn't be too hard).

What is a bigger concern with this is we want optimization to only happen once, as it may take a bit of time. But, our pipelines are currently immutable so this would happen every time you try applying it to any data.

etrain:
I'm quoting you on the first one when you become a rich and famous computer scientist. ;)

On the second part - i'm not terribly worried about this. We can build a catalog of pipelines and store them (and their optimized counterparts) somewhere, or we can add optimized pipeline as internal state (an unassigned var in the constructor). Important details but ones we can figure out. For now, optimization is a short procedure and one that we can afford to "waste" time doing multiple times.

The important part here is we want a clean separation of user-level concerns and system-level concerns.

tomerk:
Do you want this to happen both for single-item "apply" and bulk-rdd "apply"'s?

etrain:
What if the implementation is just = 
```lazy val fittedPipe =

tomerk:
I'm pretty sure you'll still end up with the infinite recursion if you do that. What I'm going to do is just call the internal execution functions of the optimized pipeline directly, as opposed to it's apply method.

PR #171, opened 2015-10-06
Apply the optimizer automatically

tomerk:
Closes issue #170 

Allows you to control what optimizer (if any) is used when executing a pipeline, defaulting to the current Optimizer object

@etrain


etrain:
This looks pretty reasonable to me. You went the "reoptimize every time" route which is probably OK for now but something we'll need to keep on our radar for the future.

Have you tested locally and verified that the examples run?

tomerk:
Yeah, if you don't want to reoptimize every time you can manually optimize and then apply the pipeline using no optimizer. And yes I've tested this locally.

PR #175, opened 2015-11-25
Fix residual mean computation

shivaram:
The previous approach of computing per-partition mean only works when number of elements per-class is equal. To fix it we now compute the sum per partition and the count per partition.

Many thanks to @tomerk and @ericmjonas for providing a test case. FWIW it wasn't being caught by the unit tests as we had equal number of elements per class in them. I'll add a unit test with uneven number of examples per class.

cc @rolloff

### DIFF
--- a/src/main/scala/nodes/learning/BlockWeightedLeastSquares.scala
+++ b/src/main/scala/nodes/learning/BlockWeightedLeastSquares.scala
@@ -162,14 +162,7 @@ object BlockWeightedLeastSquaresEstimator extends Logging {
       mat(*, ::) :- jointLabelMean
     }.cache().setName("residual")
 
-    val residualSumCount = MLMatrixUtils.treeReduce(residual.map { mat =>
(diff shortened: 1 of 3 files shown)


etrain:
This looks good, thanks @shivaram ! Do you want me to wait for a unit test to merge?

shivaram:
Yeah I'll write one -- it'll probably be easier / better if I refactor this as a utils function as well

shivaram:
I created `computeMean` in MatrixUtils and added a test case for that.

etrain:
And the unit test for the solver you mention in the PR description goes here or in `ml-matrix`?

shivaram:
It should go here but its more tricky to add -- we need to setup another test case and either verify it with Matlab or the unweighted solver or something like that. So I thought it'd be easier to test for just the `computeMean` here.

etrain:
Ok - merging this and creating an issue for you to add such a unit test.

Issue #177, opened 2015-11-25
Fix matrixToRowArray


etrain:
Need a fix for `matrixToRowArray` from @tomerk 's `keystone-vw-experiments` branch , which affects CosineRandomFeatures and perhaps other pipelines.

PR #178, opened 2015-11-25
fix for issue #177

tomerk:
fix for issue #177

### DIFF
--- a/src/main/scala/utils/MatrixUtils.scala
+++ b/src/main/scala/utils/MatrixUtils.scala
@@ -20,6 +20,8 @@ object MatrixUtils extends Serializable {
    */
   def matrixToRowArray[T : ClassTag](mat: DenseMatrix[T]): Array[DenseVector[T]] = {
     val matT = mat.t
+    // The explicit copy of the vector is necessary because otherwise Breeze slices
+    // lazily, leading to inflated serialization size (A serious issue w/ spark)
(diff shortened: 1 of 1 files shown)


tomerk:
@etrain

src/main/scala/utils/MatrixUtils.scala
```diff
@@ -18,9 +18,9 @@ object MatrixUtils extends Serializable {
    * @param mat Input matrix.
    * @return Array of rows.
    */
-  def matrixToRowArray[T](mat: DenseMatrix[T]): Array[DenseVector[T]] = {
+  def matrixToRowArray[T : ClassTag](mat: DenseMatrix[T]): Array[DenseVector[T]] = {
     val matT = mat.t
-    (0 until mat.rows).toArray.map(matT(::, _))
+    (0 until mat.rows).toArray.map(x => DenseVector(matT(::, x).toArray))
```
shivaram: Just curious: Does `.copy` or `.toDenseVector` do the same thing ?

Also would be good to add a comment here on why we are doing this

tomerk: I'm unsure whether .copy and .toDenseVector do the same thing. But I'll add a comment to make it clear why we're doing this.

shivaram:
LGTM. Thanks @tomerk

Issue #181, opened 2015-12-08
Scala Fisher Vector / GMM


tomerk:
(rename from Bens** to Scala**)

etrain:
Closed by #191

Issue #174, opened 2015-11-19
Make Scala GMM faster


tomerk:
Here's your reminder @shivaram

shivaram:
@tomerk do you know if the diff I had for this is available in your keystone repo ?

shivaram:
Never mind - I found it. I'll make a PR for this today

PR #191, opened 2015-12-17
Added the Scala KMeans++, GMM, and Fisher Vectors

tomerk:
for issue #181, and paves the way for issue #174


src/main/scala/nodes/learning/GaussianMixtureModel.scala
```diff
@@ -16,11 +17,16 @@ import workflow.{Transformer, Estimator}
  * @param variances Cluster variances (diagonal)
  * @param weights Cluster weights.
  */
-class GaussianMixtureModel(
-    val means: DenseMatrix[Double],
-    val variances: DenseMatrix[Double],
-    val weights: DenseVector[Double])
-  extends Transformer[DenseVector[Double],DenseVector[Double]] with Logging {
+case class GaussianMixtureModel(
```
etrain: why is this now a case class?

src/main/scala/nodes/learning/GaussianMixtureModel.scala
```diff
@@ -16,11 +17,16 @@ import workflow.{Transformer, Estimator}
...
+    means: DenseMatrix[Double],
+    variances: DenseMatrix[Double],
+    weights: DenseVector[Double],
+    weightThreshold: Double = 1e-4)
+  extends Transformer[DenseVector[Double],DenseVector[Double]] {
+
+  private val gmmMeans = means.t
```
etrain: what's the reason for wanting to transpose? let's decide on a convention for the constructor.

src/main/scala/nodes/learning/GaussianMixtureModel.scala
tomerk: I like to have the transformers as case classes for consistency.

src/main/scala/nodes/learning/GaussianMixtureModel.scala
tomerk: hmm good question. I think when I initially wrote this the matlab code I was translating and the native gmm code expected the transpose of each other as inputs, and we wanted them to work the same given the same input.

etrain: Agree with your logic - but we should have a declared convention. What's tricky here is that I *think* that part of the reason we did examples in columns here is because the SIFTs are column-oriented (that is, examples in each column) and therefore so are the PCAs and the GMMs. I am actually inclined to say we should just address this now, but it could have lots of downstream impacts (including on performance) so we should be careful and benchmark before we do too much.

src/main/scala/nodes/learning/GaussianMixtureModel.scala
```diff
@@ -29,11 +35,57 @@ class GaussianMixtureModel(
...
 
   /**
-   * For now this is unimplemented. It should return the soft assignment to each cluster.
+   * Returns (thresholded) assignments to each cluster.
    * @param in A Vector
-   * @return The soft assignments of the vector according to the mixture model.
+   * @return The thresholded assignments of the vector according to the mixture model.
    */
-  override def apply(in: DenseVector[Double]): DenseVector[Double] = ???
+  def apply(in: DenseVector[Double]): DenseVector[Double] = {
```
etrain: woo, an implementation, thanks!

src/main/scala/nodes/learning/GaussianMixtureModel.scala
```diff
@@ -29,11 +35,57 @@ class GaussianMixtureModel(
...
+    val numSamples = X.rows
+    val numFeatures = X.cols
+    val XSq = X :* X
+
+    /*
+    compute the squared malhanobis distance for each gaussian.
+    sq_mal_dist(i,j) || x_i - mu_j||_Lambda^2.  I am the master of
+    computing Euclidean distances without for loops.
+    */
+    val sqMahlist = (XSq * gmmVars.map(0.5 / _).t) - (X * (gmmMeans :/ gmmVars).t) + (DenseMatrix.ones[Double](numSamples, 1) * (sum(gmmMeans :* gmmMeans :/ gmmVars, Axis._1).t :* 0.5))
```
etrain: this line is super long

src/main/scala/nodes/learning/GaussianMixtureModel.scala
```diff
@@ -29,11 +35,57 @@ class GaussianMixtureModel(
...
+
+    // compute the log likelihood of the approximate posterior
+    val llh = DenseMatrix.ones[Double](numSamples, 1) * (-0.5 * numFeatures * math.log(2 * math.Pi) - 0.5 * sum(log(gmmVars), Axis._1).t + log(gmmWeights)) - sqMahlist
```
etrain: same here

src/main/scala/nodes/learning/GaussianMixtureModel.scala
```diff
@@ -29,11 +35,57 @@ class GaussianMixtureModel(
...
+
+    /*
+    if we make progress, update our pseudo-likelihood for the E-step.
+    by shifting the llh to be peaked at 0, we avoid nasty numerical
+    overflows.
+    */
+    llh(::, *) -= max(llh(*, ::))
+    exp.inPlace(llh)
+    llh(::, *) :/= sum(llh, Axis._1)
+    var q = llh
```
etrain: i don't get it - why are we creating a new variable here?

src/main/scala/nodes/learning/GaussianMixtureModel.scala
```diff
@@ -29,11 +35,57 @@ class GaussianMixtureModel(
...
+    */
+    q = q.map(x => if (x > weightThreshold) x else 0.0)
+    q(::, *) :/= sum(q, Axis._1)
+
+    q
+  }
+
+  override def apply(in: RDD[DenseVector[Double]]): RDD[DenseVector[Double]] = {
+    in.mapPartitions { partition =>
+      val assignments = apply(MatrixUtils.rowsToMatrix(partition))
```
etrain: This is fine but we should be a little careful about using this pattern everywhere. The price of doing a copy to a row matrix and then copying back when the operation is done could end up exceeding the time saved in the math. But, since your single-item implementation is doing the same basic thing, it probably won't be any faster.

src/main/scala/nodes/learning/KMeansPlusPlus.scala
```diff
@@ -0,0 +1,149 @@
...
+    apply(in.asDenseMatrix).flatten()
+  }
+
+  def apply(in: DenseMatrix[Double]): DenseMatrix[Double] = {
+    val XSqNormHlf: DenseVector[Double] = sum(in :* in, Axis._1) / 2.0
+    /* compute the distance to all of the centers and assign each point to its nearest center. */
+    val sqDistToCenters = in * means.t
+    sqDistToCenters :*= -1.0
+    sqDistToCenters(::, *) += XSqNormHlf
+    sqDistToCenters(*, ::) += (sum(means :* means, Axis._1) :*= 0.5)
```
etrain: this is super-minor, but here you in place multiply by 0.5 and above you divide by 2.0 - can we just do one for consistency?

src/main/scala/nodes/learning/KMeansPlusPlus.scala
```diff
@@ -0,0 +1,149 @@
...
+
+    val nearestCenter = argmin(sqDistToCenters(*, ::))
```
etrain: Could you add a few comments about the shape of things and why this works? Basically sqDistToCenters is numExamples by numCenters and you're using the broadcast semantics to get a big vector of closest centers for each example, right?

src/main/scala/nodes/learning/KMeansPlusPlus.scala
```diff
@@ -0,0 +1,149 @@
...
+
+    // reuse the previous (potentially large) matrix to keep memory usage down
+    val centerAssign = sqDistToCenters
+    var row: Int = 0
+    while (row < in.rows) {
+      var col: Int = 0
+      while (col < means.rows) {
+        centerAssign(row, col) = 0.0
```
etrain: wait - you need a loop for this? you can't just do something like `centerAssign = 0.0` or `centerAssign(*,*) = 0.0`?

etrain: actually - no, i'm confused by what this is doing - the point of this is to build a binary matrix indicating where all the centers are?

src/main/scala/nodes/learning/KMeansPlusPlus.scala
```diff
@@ -0,0 +1,149 @@
...
+
+    val centers = Array.fill(numMeans)(0)
+    centers(0) = Multinomial(DenseVector.fill(numSamples, 1.0/numSamples)).draw()
+
+    var curSqDistanceToClusters: DenseVector[Double] = null
+    var k = 0
+    while (k < (numMeans - 1)) {
+      val curCenter = X(centers(k), ::)
+      val curCenterNorm = norm(curCenter, 2)
+      // slick vectorized code to compute the distance to the current center
```
etrain: :+1:

src/main/scala/nodes/learning/KMeansPlusPlus.scala
```diff
@@ -0,0 +1,149 @@
...
+      val sqDistToNewCenter = (XSqNormHlf - (X * curCenter.t)) += (0.5 * curCenterNorm * curCenterNorm)
```
etrain: do we really want `+=` here?

src/main/scala/nodes/learning/KMeansPlusPlus.scala
```diff
@@ -0,0 +1,149 @@
...
+        row += 1
+      }
+
+      val assignMass = sum(centerAssign, Axis._0).toDenseVector
+      kMeans = centerAssign.t * X
+      kMeans(::, *) :/= assignMass
+
+      if (iter > 0) {
+        costImproving = (curCost(iter - 1) - curCost(iter)) >= stopTolerance * math.abs(curCost(iter - 1))
+        println("Iteration: " + iter + " current cost " + curCost(iter) + " imp " + costImproving)
```
etrain: let's make this a `logInfo` or a `logDebug`

src/main/scala/nodes/learning/ScalaFisherVector.scala
```diff
@@ -0,0 +1,44 @@
...
+    val q = gmm.apply(x.t) // numSiftDescriptors x K
+
+    /* here is the Fisher Vector in all of its beauty.  This is directly
+    from the FV survey by Sanchez et al: */
+    val s0 = mean(q, Axis._0).toDenseVector // 1 x K, but really K x 1 because it's a dense vector
+    val s1 = (x * q) :/= nDesc // D x K
+    val s2 = ((x :* x) * q) :/= nDesc // D x K
+
+    val fv1 = (s1 - gmmMeans * diag(s0)) :/ (sqrt(gmmVars) * diag(sqrt(gmmWeights)))
+    val fv2 = (s2 - (gmmMeans * 2.0 :* s1) + (((gmmMeans :* gmmMeans) - gmmVars)*diag(s0))) :/ (gmmVars * diag(sqrt(gmmWeights :* 2.0)))
```
etrain: long line

src/main/scala/nodes/learning/ScalaGMMEstimator.scala
```diff
@@ -0,0 +1,141 @@
...
+import pipelines.Logging
+import utils.MatrixUtils
+import workflow.Estimator
+
+/**
+ * Fit a Gaussian Mixture model to Data.
+ *
+ * @param k Number of centers to estimate.
+ */
+case class ScalaGMMEstimator(k: Int, maxIterations: Int = 100, minClusterSize: Int = 40, stopTolerance: Double = 1e-4, weightThreshold: Double = 1e-4, smallVarianceThreshold: Double = 1e-2, absoluteVarianceThreshold: Double = 1e-9) extends Estimator[DenseVector[Double], DenseVector[Double]] with Logging {
```
etrain: Hmm... can we avoid putting Scala in the name of the node? Why not just `GMMEstimator` the C one is in the `external` namespace anyway, isn't it? (If not, maybe it should be?)

src/main/scala/nodes/learning/ScalaGMMEstimator.scala
```diff
@@ -0,0 +1,141 @@
...
+    fit(samples.collect)
+  }
+
+  /**
+   * Fit a Gaussian mixture model with `k` centers to a sample array.
+   *
+   * @param samples Sample Array - all elements must be the same size.
+   * @return A Gaussian Mixture Model.
+   */
+  def fit(samples: Array[DenseVector[Double]]): GaussianMixtureModel = {
```
etrain: lots of long lines in this function.

src/main/scala/nodes/learning/ScalaGMMEstimator.scala
```diff
@@ -0,0 +1,141 @@
...
+      computing Euclidean distances without for loops.
+      */
+      val sqMahlist = (XSq * gmmVars.map(0.5 / _).t) - (X * (gmmMeans :/ gmmVars).t) + (DenseMatrix.ones[Double](numSamples, 1) * (sum(gmmMeans :* gmmMeans :/ gmmVars, Axis._1).t :* 0.5))
+
+      // compute the log likelihood of the approximate posterior
+      val llh = DenseMatrix.ones[Double](numSamples, 1) * (-0.5 * numFeatures * math.log(2 * math.Pi) - 0.5 * sum(bLog(gmmVars), Axis._1).t + bLog(gmmWeights)) - sqMahlist
+
+      /*
+      compute the log likelihood of the model using the incremental
+      approach suggested by the Xerox folks.  The key thing here is that
```
etrain: We probably want to reference the code/paper in the ScalaDoc above.

src/test/scala/nodes/learning/KMeansPlusPlusSuite.scala
```diff
@@ -0,0 +1,94 @@
...
+
+import breeze.linalg._
+import org.apache.spark.SparkContext
+import org.scalatest.FunSuite
+import pipelines._
+import utils.{MatrixUtils, Stats}
+
+class KMeansPlusPlusSuite extends FunSuite with LocalSparkContext with Logging {
+
+  test("K-Means++ Single Center") {
```
etrain: I weirded out by the 1-center unit test, but ok :)

src/test/scala/nodes/learning/ScalaGMMSuite.scala
```diff
@@ -0,0 +1,102 @@
...
+      DenseVector[Double](1.0, 1.0, 0.0)
+    ))
+
+    val centers = Set(
+      DenseVector[Double](1.0, 2.0, 0.0),
+      DenseVector[Double](1.0, 3.0, 6.0)
+    )
+
+    val gmm = ScalaGMMEstimator(k, minClusterSize = 1).fit(data)
+    // TODO: TEST Means & Variances!
```
etrain: maybe we can do this test?

src/test/scala/nodes/learning/ScalaGMMSuite.scala
```diff
@@ -0,0 +1,102 @@
...
+
+    /*val fitCenters = MatrixUtils.matrixToRowArray(kMeans.means).toSet
+    assert(fitCenters === centers )
+
+    val kMeans5 = KMeansPlusPlusEstimator(k, maxIterations = 5).fit(data)
+    val fitCenters5 = MatrixUtils.matrixToRowArray(kMeans5.means).toSet
+    assert(fitCenters5 === centers )
+
+    val out = kMeans.apply(data).collect()*/
+    logInfo("sup")
```
etrain: nm, you?

src/test/scala/nodes/learning/ScalaGMMSuite.scala
```diff
@@ -0,0 +1,102 @@
...
+    val center = DenseVector[Double](1.0, 3.0, 4.0).asDenseMatrix
+    val gmm = ScalaGMMEstimator(k, minClusterSize = 1).fit(data)
+
+    assert(gmm.means.t === center)
+
+    // TODO: TEST Variances!
+    gmm.apply(data)
+  }
+
+  test("GMM Two Centers") {
```
etrain: One thing to do here is actually generate the data as a mixture of Gaussians with a known mean/variance structure and then see if the model we recover is close to the data-generating model.

etrain:
This looks pretty good - most of my comments were on naming and comments. The unit test for GMM should be a little more robust, and I left an idea about how to do that. There's a minor but pretty crucial note about the use of `+=` in there, and I'd like to take another pass with my "how could in-place ops be breaking things" hat on before we merge this.

src/test/scala/nodes/learning/KMeansPlusPlusSuite.scala
tomerk: Gotta test the base case :)

src/main/scala/nodes/learning/KMeansPlusPlus.scala
tomerk: This was done at some point to reduce memory usage via an in-place addition.

src/main/scala/nodes/learning/KMeansPlusPlus.scala
tomerk: Yeah so if I recall correctly, this is basically just building a binary matrix indicating which center the point is assigned to. At some point I had some pretty non-loop code to do it, but the pretty code was allocating too much memory, so I switched to this loop code.

tomerk: I'll add a comment.

src/main/scala/nodes/learning/KMeansPlusPlus.scala
tomerk: sure

src/main/scala/nodes/learning/ScalaGMMEstimator.scala
tomerk: What is the paper again?

src/main/scala/nodes/learning/GaussianMixtureModel.scala
tomerk: Hmm okay so should I leave the convention as it is for now (each gaussian mean in `means` and gaussian var in `vars` is a column) and then just add documentation declaring the convention?

src/main/scala/nodes/learning/ScalaGMMEstimator.scala
tomerk: The C one is `nodes.learning.GaussianMixtureModelEstimator`
I'm fine moving that around and/or renaming `ScalaGMMEstimator`, just let me know what you'd like each to be.

src/main/scala/nodes/learning/GaussianMixtureModel.scala
tomerk: I'm guessing some cruft from the line-by-line matlab -> scala translation thing.

src/main/scala/nodes/learning/KMeansPlusPlus.scala
tomerk: sure, will do!

src/test/scala/nodes/learning/ScalaGMMSuite.scala
tomerk: whoops, forgot about that!

src/main/scala/nodes/learning/ScalaGMMEstimator.scala
tomerk: Fixed

src/main/scala/nodes/learning/KMeansPlusPlus.scala
tomerk: Comments added

src/main/scala/nodes/learning/KMeansPlusPlus.scala
tomerk: fixed

src/main/scala/nodes/learning/GaussianMixtureModel.scala
tomerk: Comments added

tomerk:
@etrain so I think the only things left are adding a reference to the GMM paper (what is the paper?)
And figuring out the naming / name spaces.

The C GMM estimator is nodes.learning.GaussianMixtureModelEstimator
I'm fine moving that around and/or renaming ScalaGMMEstimator, just let me know what you'd like each to be.

etrain:
Let's put the let's have a `nodes.learning.external.GaussianMixtureModelEstimator` which is the C thing (and be sure to update all existing pipelines to use this) and then name the scala one just `nodes.learning.GaussianMixtureModelEstimator`. We've got this convention on the featurizers as well, so it makes sense to me organizationally.

etrain:
Appendix B of the following paper has guidelines for training the GMM:

Jorge Sanchez, Florent Perronnin, Thomas Mensink, Jakob Verbeek. Image Classification with
the Fisher Vector: Theory and Practice. International Journal of Computer Vision, Springer
Verlag, 2013, 105 (3), pp.222-245. <10.1007/s11263-013-0636-x>. <hal-00830491v2>

tomerk:
Okay @etrain how's this looking?

etrain:
It doesn't look like you updated the existing pipelines that use the GMM/FV nodes to point to the external namespace?

tomerk:
Ah, I can see where the confusion comes from. So, the pipelines never use the GaussianMixtureModelEstimator directly, they only use the GMMFisherVectorEstimator (already in external), which I have now changed to point at the external GMM Estimator. And, the ones that use FisherVector were already pointing at the external.FisherVector node.

etrain:
Ah, I see - sorry, I should have verified :)

LGTM!

Issue #184, opened 2015-12-08
Hashing trick


tomerk:
Add in HashingTF and NGramsHashingTF nodes from my keystone-vw-experiments

etrain:
Closed by #193

PR #193, opened 2015-12-18
Added hashing trick nodes

tomerk:
issue #184

### DIFF
--- a/src/main/scala/nodes/nlp/NGramsHashingTF.scala
+++ b/src/main/scala/nodes/nlp/NGramsHashingTF.scala
@@ -13,7 +13,8 @@ import scala.collection.mutable
  *
  * It computes a rolling MurmurHash3 instead of fully constructing the n-grams, making
  * it more efficient than using [[NGramsFeaturizer]] followed by [[HashingTF]], although
- * it should return the exact same feature vector.
+ * it should return the exact same feature vector. The MurmurHash3 methods are copied from
+ * [[scala.util.hashing.MurmurHash3]]
  *
  * Individual terms are hashed using Scala's `.##` method. We may want to convert to MurmurHash3 for strings,
  * as discussed for Spark's ML Pipelines in https://issues.apache.org/jira/browse/SPARK-10574
--- a/src/main/scala/nodes/nlp/NGramsHashingTF.scala
+++ b/src/main/scala/nodes/nlp/NGramsHashingTF.scala
@@ -1,0 +1,117 @@
+package nodes.nlp
+
+import java.lang.Integer.{rotateLeft => rotl}
+
+import breeze.linalg.SparseVector
+import workflow.Transformer
+
+import scala.collection.mutable
+
+/**
+ * Converts the n-grams of a sequence of terms to a sparse vector representing their frequencies,
+ * using the hashing trick: https://en.wikipedia.org/wiki/Feature_hashing
+ *
+ * It computes a rolling MurmurHash3 instead of fully constructing the n-grams, making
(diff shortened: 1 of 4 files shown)


src/main/scala/nodes/nlp/NGramsHashingTF.scala
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Converts the n-grams of a sequence of terms to a sparse vector representing their frequencies,
+ * using the hashing trick: https://en.wikipedia.org/wiki/Feature_hashing
+ *
+ * It computes a rolling MurmurHash3 instead of fully constructing the n-grams, making
+ * it more efficient than using [[NGramsFeaturizer]] followed by [[HashingTF]], although
+ * it should return the exact same feature vector.
+ *
+ * Individual terms are hashed using Scala's `.##` method. We may want to convert to MurmurHash3 for strings,
```
etrain: Is this comment still relevant?

src/main/scala/nodes/nlp/NGramsHashingTF.scala
```diff
@@ -0,0 +1,117 @@
...
+    case _ =>
+  }
+
+  final val seqSeed = "Seq".hashCode
+
+  /** Mix in a block of data into an intermediate hash value. */
+  final def mix(hash: Int, data: Int): Int = {
+    var h = mixLast(hash, data)
+    h = rotl(h, 13)
+    h * 5 + 0xe6546b64
```
etrain: Can we get some comments about where these constants come from?

src/test/scala/nodes/nlp/NGramsHashingTFSuite.scala
```diff
@@ -0,0 +1,59 @@
...
+  test("NGramsHashingTF 1 to 1") {
+    val dims = 40000
+
+    val testDatum = "this sentence is a sentence is the some there some then there some".split(" ")
+    val ngrams = NGramsFeaturizer(1 to 1).apply(testDatum)
+    val tfVector = HashingTF(dims).apply(ngrams)
+
+    val ngramsHashingTFVector = NGramsHashingTF(1 to 1, dims).apply(testDatum)
+
+    // Assert that the NGramsHashingTF node returns the same output as first getting n-grams then hashing
```
etrain: yes, i like this test!

etrain:
This looks good to me - if you could clean up the comments and add a direct reference to the Murmur Hash 3 algorithm. Is there a reason we're reimplementing rather than linking to the one in the scala stdlib? 

http://www.scala-lang.org/api/2.10.4/#scala.util.hashing.MurmurHash3$

tomerk:
Yeah the code (and constants) come directly from the scala stdlib, but the scala stdlib marked these rolling hash methods as `private[hashing]`. I could make a util in the hashing namespace that exposes these methods to us, which may be preferable to the code & constants copying I've done here.

src/main/scala/nodes/nlp/NGramsHashingTF.scala
tomerk: Yeah, the terms' hash values are combined together using MurmurHash3 to form the n-gram hashes, but the individual terms are still hashed using the `.##` method.

etrain:
Copy/pasting is fine, then, I guess. Let's just leave a reference to where it came from in the code?

tomerk:
Okay @etrain I added a reference to the scala stdlib class.

etrain:
LGTM, merging this!

Issue #179, opened 2015-12-08
PCA Nodes


tomerk:
Introduce from my experiments branch:
Distributed PCA
Approximate Local PCA
Approximate Distributed PCA
(Requires moving over TruncatedSVD and ColumnPartitionedMatrix)

PR #194, opened 2015-12-18
Pca nodes

tomerk:
Issue #179 
This includes the distributed PCA node and the local approximate PCA node and unit tests.

### DIFF
--- a/src/main/scala/nodes/learning/ApproximatePCA.scala
+++ b/src/main/scala/nodes/learning/ApproximatePCA.scala
@@ -59,9 +59,9 @@ class ApproximatePCAEstimator(dims: Int, q: Int = 10, p: Int = 5)
     val pca = convert(usvt.Vt.t, Float)
     logInfo(s"shape of pca (${pca.rows},${pca.cols}")
 
-    PCAEstimator.enforceMatlabPCASignConvention(pca)
+    val matlabConventionPCA = PCAEstimator.enforceMatlabPCASignConvention(pca)
 
     // Return a subset of the columns.
-    pca(::, 0 until k)
+    matlabConventionPCA(::, 0 until k)
   }
 }
--- a/src/main/scala/nodes/learning/ApproximatePCA.scala
+++ b/src/main/scala/nodes/learning/ApproximatePCA.scala
@@ -11,9 +11,14 @@ import pipelines.Logging
 import workflow.Estimator
 
 /**
- * Created by tomerk11 on 12/16/15.
+ * Approximately estimates a PCA model for dimensionality reduction based on an input dataset.
+ *
+ * @param dims Dimensions to reduce input dataset to.
+ * @param q The number of iterations to use
+ * @param p The amount of padding to add beyond dims for the calculations
  */
-class ApproximatePCAEstimator(dims: Int, q: Int = 10, p: Int = 5) extends Estimator[DenseVector[Float], DenseVector[Float]] with Logging {
+class ApproximatePCAEstimator(dims: Int, q: Int = 10, p: Int = 5)
+    extends Estimator[DenseVector[Float], DenseVector[Float]] with Logging {
 
   /**
    * Adapted from the "PCA2" matlab code given in appendix B of this paper:
@@ -24,15 +29,15 @@ class ApproximatePCAEstimator(dims: Int, q: Int = 10, p: Int = 5) extends Estima
    */
   def fit(samples: RDD[DenseVector[Float]]): PCATransformer = {
     val samps = samples.collect.map(_.toArray)
-    val dataMat: DenseMatrix[Double] = convert(DenseMatrix(samps:_*), Double)
-    new PCATransformer(convert(approximatePCA(dataMat, dims, q, p), Float))
+    val dataMat: DenseMatrix[Float] = DenseMatrix(samps:_*)
+    new PCATransformer(approximatePCA(dataMat, dims, q, p))
   }
 
-  def approximatePCA(data: DenseMatrix[Double], k: Int, q: Int = 10, p: Int = 5): DenseMatrix[Double] = {
+  def approximatePCA(data: DenseMatrix[Float], k: Int, q: Int = 10, p: Int = 5): DenseMatrix[Float] = {
     //This algorithm corresponds to Algorithms 4.4 and 5.1 of Halko, Martinsson, and Tropp, 2011.
     //According to sections 9.3 and  9.4 of the same, Ming Gu argues for exponentially fast convergence.
 
-    val A = data
+    val A = convert(data, Double)
     val d = A.cols
 
     val l = k + p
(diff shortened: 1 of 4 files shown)


tomerk:
Do we also want the approximate distributed PCA node or does it's utility over the other three nodes seem too limited for the most part?

etrain:
Let's do the approximate PCA node while we're at it - it's useful in cases where data is extremely large and pretty low-rank, which is actually something I could see happening in practice (think huge sparse vectors that contain little info/lossily compress really well).

src/main/scala/nodes/learning/ApproximatePCA.scala
```diff
@@ -0,0 +1,73 @@
...
+import edu.berkeley.cs.amplab.mlmatrix.util.QRUtils
+import org.apache.spark.rdd.RDD
+import org.netlib.util.intW
+import pipelines.Logging
+import workflow.Estimator
+
+/**
+ * Created by tomerk11 on 12/16/15.
+ */
+class ApproximatePCAEstimator(dims: Int, q: Int = 10, p: Int = 5) extends Estimator[DenseVector[Float], DenseVector[Float]] with Logging {
```
etrain: too big

src/main/scala/nodes/learning/ApproximatePCA.scala
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Adapted from the "PCA2" matlab code given in appendix B of this paper:
+   *    https://www.cs.princeton.edu/picasso/mats/PCA-Tutorial-Intuition_jp.pdf
+   *
+   * @param samples A sample of features to be reduced. Often O(1e6). Logically row-major.
+   * @return A PCA Matrix which will perform dimensionality reduction when applied to a data matrix.
+   */
+  def fit(samples: RDD[DenseVector[Float]]): PCATransformer = {
+    val samps = samples.collect.map(_.toArray)
+    val dataMat: DenseMatrix[Double] = convert(DenseMatrix(samps:_*), Double)
```
etrain: Why not `MatrixUtils.rowsToMatrix`?

src/main/scala/nodes/learning/ApproximatePCA.scala
etrain: Can you add a scaladoc string explaining what each value is?

src/main/scala/nodes/learning/ApproximatePCA.scala
```diff
@@ -0,0 +1,73 @@
...
+    val B = Q.t * A //cpu: l*n*d, mem: l*d
+    val usvt = svd.reduced(B) //cpu: l*d^2, mem: l*d
+    val pca = usvt.Vt.t
+    logInfo(s"shape of pca (${pca.rows},${pca.cols}")
+
+    // Mimic matlab
+    // Enforce a sign convention on the coefficients -- the largest element in
+    // each column will have a positive sign.
+
+    val colMaxs = max(pca(::, *)).toArray
```
etrain: I wonder if this should be factored out into a separate function since we do this in several places.

src/main/scala/nodes/learning/DistributedPCA.scala
```diff
@@ -0,0 +1,68 @@
...
+ * @param dims Dimensions to reduce input dataset to.
+ */
+class DistributedPCAEstimator(dims: Int) extends Estimator[DenseVector[Float], DenseVector[Float]] with Logging {
+
+  /**
+   * Adapted from the "PCA2" matlab code given in appendix B of this paper:
+   *    https://www.cs.princeton.edu/picasso/mats/PCA-Tutorial-Intuition_jp.pdf
+   *
+   * @param samples Features to be reduced. Logically row-major.
+   * @return A PCA Matrix which will perform dimensionality reduction when applied to a data matrix.
```
etrain: This is not quite right. The result is a PCA Transformer.

src/test/scala/nodes/learning/PCASuite.scala
```diff
@@ -79,4 +81,49 @@ class PCATransformerSuite extends FunSuite with LocalSparkContext with Logging {
...
+    val dimRed = 5
+
+    // Generate a random Gaussian matrix.
+    val gau = new Gaussian(0.0, 1.0)
+    val randMatrix = new DenseMatrix(matRows, matCols, gau.sample(matRows*matCols).toArray)
+
+    // Parallelize and estimate the PCA.
+    val data = sc.parallelize(MatrixUtils.matrixToRowArray(randMatrix).map(x => convert(x, Float)))
+
+    val pcaApprox = new ApproximatePCAEstimator(dimRed, q = 500).fit(data)
```
etrain: Do we have to run for 500 iterations to get sufficient accuracy? If so, something is not quite right..

etrain:
Looks mostly OK. Let's factor out the Matlab conversion thing and maybe dig into why ApproximatePCA isn't getting "close" to the real thing with fewer iterations. At 500 iters the algorithm really isn't worth running.

src/main/scala/nodes/learning/ApproximatePCA.scala
tomerk: Hmm I would guess you got this code from `PCAEstimator` which was probably written before we had the `MatrixUtils.rowsToMatrix` helper method. I can change it both here and in `PCAEstimator`

src/main/scala/nodes/learning/ApproximatePCA.scala
tomerk: An "enforce matlab sign convention" method?

src/main/scala/nodes/learning/DistributedPCA.scala
tomerk: Sure, will update the doc!

src/test/scala/nodes/learning/PCASuite.scala
tomerk: Yeah so I can lower the number of iterations to ~10 or so with no problem if I set dimRed to matCols. But at a low dimRed the results get shockingly different at low iteration counts.

tomerk: Hmm, looks like if I increase p when I lower dimReds it'll also work better (but I have no sense of what implications that has on performance / etc.)

etrain: Basically the runtime scales with q(k+p)^2 - so a really big p is a bad
idea, but if it's small (say 5 or 10) it's probably ok.

etrain:
A minor nit about the way the matlab sign stuff is written, and I guess we still want a good test for the approximate stuff?

tomerk:
Sure, I can fix the matlab sign stuff to not be an in-place op.
As for the approximate stuff, I can't seem to get the results to be close to the output of the local PCA. (I regularly see as high as ~20-40% error on average per cell value)

etrain:
Hmm.. 20-40% is high but not insane, and could just be a function of how we're running things. I'll see if I can come up with a better test later today.

tomerk:
@etrain just let me know when you know of a better test you want me to implement!

tomerk:
That may have been the first time I've ever done a pull request on a pull request. @etrain how's this looking now?

etrain:
After digging in, I think the approximate algorithm *is* fulfilling its contract in 5-10 iterations, which is great, and with minimal padding (again, 5 should be plenty). The additional tests verify that the algorithmic guarantees are holding, and that from a functional level, the node is doing what we want (driving covariance to be diagonal).

Issue #200, opened 2016-01-06
Scala GMM and EncEval GMM don't give the same results


jrabary:
I'm trying to compare the results of the two implementations of GMM estimator in keystoneml and they are very differents. As sample data, I use the `gmm_data.txt` available in spark source.
The EncEval version give results similar to scikit-learn GMM.

  
~~~
import nodes.learning.GaussianMixtureModelEstimator
import nodes.learning.external.{GaussianMixtureModelEstimator => EGMM}
import org.apache.spark.mllib.clustering.{GaussianMixtureModel => SGMM, GaussianMixture}
import org.apache.spark.mllib.linalg.Vectors
import org.apache.spark.{SparkConf, SparkContext}
import breeze.linalg.{DenseVector => BDV}


object GMM {

  def main (args: Array[String]) {
    val conf = new SparkConf()
      .setAppName("Extract Face Features")
      .setIfMissing("spark.master", "local[4]")
    val sc = new SparkContext(conf)
    val data = sc.textFile("spark/data/mllib/gmm_data.txt")
    val parsedData = data.map(s => BDV[Double](s.trim.split(' ').map(_.toDouble))).cache()

    val gmmEstimator = GaussianMixtureModelEstimator(2, 
        maxIterations = 100000,stopTolerance = 0d,
        smallVarianceThreshold = 0d)
    val gmmModel = gmmEstimator.fit(parsedData)

    val gmmEstimator2 = new EGMM(2)
    val gmmModel2 = gmmEstimator2.fit(parsedData)

    println("===================")
    println(gmmModel.means)
    println(gmmModel2.means)
    println("-------")
    println(gmmModel.variances)
    println(gmmModel2.variances)
    println("-----------")
    println(gmmModel.weights)
    println(gmmModel2.weights)

    sc.stop()
  }
}
~~~

etrain:
Thanks for the bug report and code example - very helpful!

There are a few differences between the two implementations, one important one is the initialization strategy. The scala version is initialized via KMeans++, while the EncEval library version is initialized uniformly at random in data domain. A quick look also inidcates that the scala version actually achieves a better log-likelihood near the end of its iterations. -3.609 for EncEval vs. -3.564 for the Scala version.

etrain:
Hey there - @tomerk and I took a look at this today.

etrain:
[...]

PR #201, opened 2016-01-07
Made the GMM Estimator take a seed and have configurable initialization

tomerk:


etrain:
Ugh, left my comments on the commit rather than the "files changed" but this looks pretty good. If you could fix the few style things I pointed out that would be great. Thanks @tomerk !

etrain:
Actually - while you're at it if you could add a test that runs on the data set I just posted to #200 and checks that the output of both implementations on that dataset are "close" to (0,0) for means and (25,1) or (1,25) for sigma, as well as close to 50/50 for the mixture weights, that would be great!

etrain:
here "close" can be liberal.

tomerk:
@etrain took care of all the above

src/main/scala/nodes/learning/GaussianMixtureModelEstimator.scala
```diff
@@ -63,13 +63,42 @@ case class GaussianMixtureModelEstimator(
...
+
+      case RANDOM_INITIALIZATION =>
+        // Random Initialization
+        val colMin = min(X(::, *)).toDenseVector
+        val colMax = max(X(::, *)).toDenseVector
+        val colRange = colMax - colMin
+
+        val rand = new RandBasis(new ThreadLocalRandomGenerator(new MersenneTwister(seed)))
+
+        val gmmMeans = DenseMatrix.rand(k, numFeatures, Uniform(0, 1)(rand))
```
etrain: nice

src/test/scala/nodes/learning/GaussianMixtureModelSuite.scala
```diff
@@ -90,6 +90,31 @@ class GaussianMixtureModelSuite extends FunSuite with LocalSparkContext with Log
...
+
+    val gmmEst = GaussianMixtureModelEstimator(k, minClusterSize = 1, seed = 0, stopTolerance = 0, maxIterations = 30)
+
+    val data = sc.parallelize(TestUtils.loadFile("gmm_data.txt"))
+    val parsedData = data.map(s => DenseVector[Double](s.trim.split(' ').map(_.toDouble))).cache()
+
+
+    val centers = DenseMatrix.zeros[Double](2, 2)
+
+    val variances = DenseMatrix((1.0, 25.0), (25.0, 1.0))
```
etrain: Hmm.. is the fit going to be deterministically in this order?

etrain:
All of this looks good to me except for the last comment about the test - or does the fixed seed just solve this?

tomerk:
The deterministic seed solves the ordering.

PR #203, opened 2016-01-13
Handle empty partitions in WeightedLeastSquares

### LINKED ISSUE #196: BlockSolve should handle empty partitions
After discussion with @tomerk it seems that it would be really useful to have the various Block Sovlers not choke on empty partitions. Empty partitions can arise in the course of cross-validation when you want to fiter your data RDD into a "train" rdd and a "test" rdd.

shivaram:
Fixes #196

### DIFF
--- a/src/main/scala/nodes/learning/BlockWeightedLeastSquares.scala
+++ b/src/main/scala/nodes/learning/BlockWeightedLeastSquares.scala
@@ -129,14 +129,19 @@ object BlockWeightedLeastSquaresEstimator extends Logging {
     }
 
     val classIdxs = labels.mapPartitions { iter =>
-      Iterator.single(iter.map(label => label.toArray.indexOf(label.max)).toSeq.distinct.head)
+      if (iter.hasNext) {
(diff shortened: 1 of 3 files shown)


shivaram:
The solver seems to run successfully both with and without groupByClasses - but the solution is weird in the test case I have. I'm trying to construct another test case, but I thought it'll be useful to get this out for  @ericmjonas to test

etrain:
@ericmjonas - maybe you can try this out on your problem and let us know if it works before we merge it in?

etrain:
This LGTM - shall we merge and revisit if it doesn't solve @ericmjonas's problem?

shivaram:
Yeah that sounds good to me. One more thing we could do is add a train/test split like thing that @ericmjonas is doing in one of our pipelines -- That way we could add a integration test that covers this ?

etrain:
I think that's a good idea - one reasonable idea would be to use one of the Solar Flares pipelines itself as an example and add it to the integration tests suite.

PR #204, opened 2016-01-13
add an image crop utility function

stephentu:
Here's a simple function to crop an image into a bounding box. I needed this for feature augmentation, and figured it was useful independently for keystone.


shivaram:
@etrain Should we make a new node for this ?

etrain:
My thoughts exactly - a "Cropper"? it's implementation can be to call into this function directly.

etrain:
I'm going to go ahead and merge this. I've created #211 to track creation of a `Cropper` but I think this is useful enough on its own.

etrain:
Thanks @stephentu!

PR #205, opened 2016-01-20
Fixing daisy feature ordering to match SIFT. Closes #186.

### LINKED ISSUE #186: Daisy Extractor Fix

etrain:
This fix transposes the order of features coming out of the Daisy Extractor so that it's a drop in replacement for the SIFT featurizer -- logically these things do different things but at least "Image Descriptors" should have a common structure. Closes #186

### DIFF
--- a/src/main/scala/nodes/images/DaisyExtractor.scala
+++ b/src/main/scala/nodes/images/DaisyExtractor.scala
@@ -187,7 +187,7 @@ class DaisyExtractor(
       }
       x += 1
     }
-    out
+    out.t //Note: to match SIFT, the output should be daisyFeatureSize x num-descriptors.
   }
 
   def normalize(x: DenseVector[Double]): DenseVector[Double] = {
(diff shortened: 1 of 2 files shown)


shivaram:
LGTM

shivaram:
Would be good to have a test comparing them somehow ? i.e. enforcing they are the same in Daisy and SIFT ?

etrain:
I can do that, but now I'm wondering if an `ImageDescriptor` class is a better data structure to be using here, rather than enforcing object structure by convention. The downside to this is that either downstream operators (e.g. PCA, GMM, FV) would need to be changed to handle this new class (probably too big for this PR) or a conversion node would have to be invented that extracts out the matrix in the right order, which feels like a hack.

I'm going to add a test for this, and I've opened #210 to address the `ImageDescriptor` problem in a future PR.

etrain:
Test case added.

shivaram:
LGTM

PR #208, opened 2016-01-20
Fixing VectorSplitter irregular block size issue and adding unit test.

### LINKED ISSUE #188: VectorSplitter should check for end of vector
The vector splitter right now assumes that blockSize is a multiple of the vector length. Instead it 
should do `min(vec.length, (blockNum + 1) * blockSize)` to figure out the end index.

etrain:
This adds a fix for the non-apply method in VectorSplitter which doesn't have a proper termination condition when the blocksize is not an exact multiple of the vector size.

Closes #188

### DIFF
--- a/src/test/scala/nodes/util/VectorSplitterSuite.scala
+++ b/src/test/scala/nodes/util/VectorSplitterSuite.scala
@@ -1,0 +1,23 @@
+package nodes.util
+
+import breeze.linalg.DenseVector
+import org.scalatest.FunSuite
+
+class VectorSplitterSuite extends FunSuite {
+  test("vector splitter") {
+    for (
+      bs <- Array(128, 256, 512, 1024, 2048);
+      mul <- 0 to 2;
+      off <- 0 to 20 by 5;
+      feats <- Array(Some(bs*mul+off), None)
+    ) {
+      val sp = new VectorSplitter(bs, feats)
+      val vec = DenseVector.zeros[Double](bs*mul+off)
+
+      val expectedSplits = (bs*mul+off)/bs + (if ((bs*mul+off)%bs == 0) 0 else 1)
+
+      assert(sp.splitVector(vec).length === expectedSplits,
+        s"True length is ${sp.splitVector(vec).length}, expected length is ${expectedSplits}")
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


src/test/scala/nodes/util/VectorSplitterSuite.scala
```diff
@@ -0,0 +1,23 @@
...
+    for (
+      bs <- Array(128, 256, 512, 1024, 2048);
+      mul <- 0 to 2;
+      off <- 0 to 20 by 5;
+      feats <- Array(Some(bs*mul+off), None)
+    ) {
+      val sp = new VectorSplitter(bs, feats)
+      val vec = DenseVector.zeros[Double](bs*mul+off)
+
+      val expectedSplits = (bs*mul+off)/bs + (if ((bs*mul+off)%bs == 0) 0 else 1)
```
shivaram: I think we use spaces between `+` and `%`s

src/test/scala/nodes/util/VectorSplitterSuite.scala
```diff
@@ -0,0 +1,23 @@
...
+
```
shivaram: The other thing to assert here is to join all the splits and make sure you get the same vector as before

shivaram:
Sorry I should have done this in the same solvers PR. Code change LGTM. Some minor test comments

etrain:
Fixed formatting issues and added a test.

src/test/scala/nodes/util/VectorSplitterSuite.scala
```diff
@@ -0,0 +1,40 @@
...
+    for (
+      bs <- Array(128, 256, 512, 1024, 2048);
+      mul <- 0 to 2;
+      off <- 0 to 20 by 5;
+      feats <- Array(Some(bs*mul + off), None)
+    ) {
+      val sp = new VectorSplitter(bs, feats)
+      val vec = rand(bs*mul + off)
+
+      val expectedSplits = (bs*mul + off) / bs + (if ((bs*mul+off) % bs == 0) 0 else 1)
```
shivaram: you can delete this line or move the assertion on the following line to the previous test case as the setup is the same for both of them

etrain:
good catch, thanks!

shivaram:
LGTM

PR #209, opened 2016-01-20
Changing default ZCA Whitener regularization parameter and adding documentation.

### LINKED ISSUE #190: Fix ZCA Regularization parameter in CIFAR pipeline
The current regularization value for the ZCAEstimator in the CIFAR pipeline is very small (1e-12)  -- we should increase it to 0.1 for this pipeline and consider increasing the default value in general.

etrain:
Adding some docs and changing the default regularization parameter. This significantly improves accuracy on e.g. the CIFAR pipeline and will in general improve accuracy when input data has very small singular values, which you'd expect to be the case in "compressible" data.

Closes #190

### DIFF
--- a/src/main/scala/nodes/learning/ZCAWhitener.scala
+++ b/src/main/scala/nodes/learning/ZCAWhitener.scala
@@ -17,7 +17,14 @@ class ZCAWhitener(val whitener: DenseMatrix[Double], val means: DenseVector[Doub
   }
 }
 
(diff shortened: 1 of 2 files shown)


shivaram:
Can you also make this configurable from the CIFAR pipeline ? i.e add a command line config to the pipeline file etc.

etrain:
Good idea! Done.

shivaram:
LGTM

PR #212, opened 2016-01-25
Adding image cropper class.

### LINKED ISSUE #211: Cropper Node
It looks like cropping will be useful particularly for data augmentation. In #204 we have an Image Crop utility function. We should create a `Cropper` node.

etrain:
This is a follow on to #204 which adds a `Cropper` transformer.

Closes #211


etrain:
Unit tests are covered by the ones already written for the equivalent `ImageUtils` method so I omitted those.

shivaram:
LGTM

PR #216, opened 2016-01-27
Adding class label indicators validation.

### LINKED ISSUE #197: ClassLabelIndicatorsFromIntLabels should error when given labels outside [0, numClasses)
It's not immediately obvious that `ClassLabelIndicatorsFromIntLabels` wants contiguous class labels, and a naive user might give it labels like {-1, +1} or simply unique keys. Throwing an error here would be great.

etrain:
Closes #197 

Added unit tests.

### DIFF
--- a/src/main/scala/nodes/util/ClassLabelIndicators.scala
+++ b/src/main/scala/nodes/util/ClassLabelIndicators.scala
@@ -8,7 +8,7 @@ import workflow.Transformer
 /**
  * Given a class label, returns a binary vector that indicates when that class is present.
  *
- * Expects labels in the range [1, numClasses).
+ * Expects labels in the range [0, numClasses) and numClasses > 1.
  *
  * @param numClasses
  */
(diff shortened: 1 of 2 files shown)


src/main/scala/nodes/util/ClassLabelIndicators.scala
```diff
@@ -21,12 +30,21 @@ case class ClassLabelIndicatorsFromIntLabels(numClasses: Int)
...
  * @param numClasses
  */
-case class ClassLabelIndicatorsFromIntArrayLabels(numClasses: Int)
+case class ClassLabelIndicatorsFromIntArrayLabels(numClasses: Int, validate: Boolean = false)
     extends Transformer[Array[Int], DenseVector[Double]] {
 
+  assert(numClasses > 1, "numClasses must be > 1.")
+
   def apply(in: Array[Int]): DenseVector[Double] = {
+    if(validate && (in.max > numClasses || in.min < 0)) {
```
shivaram: I think this should be `>= numClasses` ?

src/main/scala/nodes/util/ClassLabelIndicators.scala
```diff
@@ -7,12 +7,21 @@ import workflow.Transformer
 
 /**
  * Given a class label, returns a binary vector that indicates when that class is present.
+ *
+ * Expects labels in the range [1, numClasses).
```
shivaram: This should be `[0, numClasses)` ?

etrain:
Good catches.

shivaram:
LGTM

Issue #198, opened 2016-01-01
cifar RandomPatchCifar crashes with segfault


stephentu:
Using keystone straight out of the box, i run into the following issue:
```
sltu@c77:/scratch/stephentu/keystone(master)$ KEYSTONE_MEM=200g ./bin/run-pipeline.sh  \
  pipelines.images.cifar.RandomPatchCifar \
  --trainLocation ./cifar_train.bin \
  --testLocation ./cifar_test.bin \
  --numFilters 1024 \
  --lambda 3000                                             
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/scratch/stephentu/keystone/target/scala-2.10/keystoneml-assembly-0.3.0-SNAPSHOT.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/scratch/stephentu/spark-1.3.1-bin-cdh4/lib/spark-assembly-1.3.1-hadoop2.0.0-mr1-cdh4.2.0.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.Log4jLoggerFactory]
16/01/01 00:00:46 INFO Cacher: CACHING 8                                                                                                   
16/01/01 00:00:46 INFO ConcretePipeline: Fitting 'StandardScaler' [10]
[Stage 2:>                                                                                                                   (0 + 48) / 48]16/01/01 00:24:55 INFO ConcretePipeline: Finished fitting 'StandardScaler' [10]                                                            
16/01/01 00:24:55 INFO Cacher: CACHING 16
16/01/01 00:24:55 INFO Cacher: CACHING 17
16/01/01 00:24:55 INFO Cacher: CACHING 19
#                                                                                                                                          
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f02a54fed96, pid=42789, tid=139431823296256
#
# JRE version: OpenJDK Runtime Environment (7.0_91-b02) (build 1.7.0_91-b02)
# Java VM: OpenJDK 64-Bit Server VM (24.91-b01 mixed mode linux-amd64 )
# Derivative: IcedTea 2.6.3
# Distribution: Ubuntu 14.04 LTS, package 7u91-2.6.3-0ubuntu0.14.04.1
[...]

PR #217, opened 2016-01-28
Adding logic to set OMP_NUM_THREADS appropriately

etrain:
Addresses #198 

Tested locally on OSX in several combinations of situations, and verified that env variables were set as expected.

### DIFF
--- a/bin/run-pipeline.sh
+++ b/bin/run-pipeline.sh
@@ -31,8 +31,6 @@ fi
 
 EXECUTOR_OMP_NUM_THREADS=${EXECUTOR_OMP_NUM_THREADS:-1}
 
-echo $OMP_NUM_THREADS $CORES $EXECUTOR_OMP_NUM_THREADS
-
 if [[ -z "$SPARK_HOME" ]]; then
   echo "SPARK_HOME is not set, running pipeline locally"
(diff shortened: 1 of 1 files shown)


bin/run-pipeline.sh
```diff
@@ -6,6 +6,32 @@ FWDIR="$(cd `dirname $0`; pwd)"
 CLASS=$1
 shift
 
+# Set OMP_NUM_THREADS on workers and driver to something appropriate.
+# This is due to OpenBLAS not handling large numbers of cores very well.
```
shivaram: Lets add a link to the keystone issue here. I'll add my openblas commit investigation to the same issue

shivaram:
Change looks fine. Can we test this on Linux on say a millennium machine before merging ?

bin/run-pipeline.sh
etrain: Will do.

etrain:
Great - I will test on millennium this AM.

etrain:
Apparrently `[[` and `-eq` don't behave as expected with strings. Once that was fixed everything seems to work fine on both linux and osx.

shivaram:
Cool. LGTM

PR #220, opened 2016-01-30
Pipeline instruction conversions

tomerk:
Introduced "Instructions" to allow easier interpretation/execution of pipelines, and utils for converting between a Pipeline DAG and Seq[Instructions]


src/main/scala/pipelines/images/imagenet/ImageNetSiftLcsFV.scala
```diff
@@ -21,7 +21,7 @@ import nodes.stats._
 import nodes.util._
 
 import utils.{Image, MatrixUtils, Stats}
-import workflow.{Optimizer, Pipeline}
+import workflow.{DefaultOptimizer, Pipeline}
```
etrain: I don't think these imports are necessary now that `pipeline.fit` calls into the optimizer directly.

src/main/scala/pipelines/images/voc/VOCSIFTFisher.scala
```diff
@@ -14,7 +14,7 @@ import nodes.util.{Cacher, ClassLabelIndicatorsFromIntArrayLabels, FloatToDouble
 import org.apache.spark.{SparkConf, SparkContext}
 import pipelines.Logging
 import scopt.OptionParser
-import workflow.Optimizer
+import workflow.DefaultOptimizer
```
etrain: same here - and in timit, amazon, and newsgroups.

src/main/scala/workflow/ConcretePipeline.scala
```diff
@@ -85,11 +87,14 @@ private[workflow] class ConcretePipeline[A, B](
...
+          val nodeDataDeps = dataDeps(node).map(x => singleDataEval(x, in))
+          transformer.transform(nodeDataDeps)
+        case delTransformer: DelegatingTransformerNode =>
+          val nodeFitDep = fitDeps(node).map(fitEstimator).get
           val nodeDataDeps = dataDeps(node).map(x => singleDataEval(x, in))
-          transformer.transform(nodeDataDeps, nodeFitDeps)
-        case _: DataNode => throw new RuntimeException(
+          nodeFitDep.transform(nodeDataDeps)
+        case _: SourceNode => throw new RuntimeException(
           "Pipeline DAG error: came across an RDD data dependency when trying to do a single item apply"
```
etrain: nit - should this say "source dependency" now that we've changed the name?

src/main/scala/workflow/ConcretePipeline.scala
```diff
@@ -119,9 +129,9 @@ private[workflow] class ConcretePipeline[A, B](
     }
   }
 
-  override def apply(in: A): B = apply(in, Some(Optimizer))
+  override def apply(in: A): B = apply(in, Some(DefaultOptimizer))
 
-  override def apply(in: RDD[A]): RDD[B] = apply(in, Some(Optimizer))
+  override def apply(in: RDD[A]): RDD[B] = apply(in, Some(DefaultOptimizer))
 
   def apply(in: A, optimizer: Option[RuleExecutor]): B = {
```
etrain: Do we still want this to be an `Option[RuleExecutor]`now that we have an `Optimizer` type?

src/main/scala/workflow/DefaultOptimizer.scala
```diff
@@ -17,7 +17,7 @@ object Optimizer extends RuleExecutor {
 object EquivalentNodeMerger extends Rule {
   def apply[A, B](plan: Pipeline[A, B]): Pipeline[A, B] = {
     val fullNodes = plan.nodes.zip(plan.dataDeps.zip(plan.fitDeps)).zipWithIndex
-    val mergableNodes = fullNodes.groupBy(_._1).mapValues(_.map(_._2)).toSeq
+    val mergableNodes = fullNodes.groupBy(_._1).mapValues(_.map(_._2)).toSeq.sortBy(_._2.min)
```
etrain: :+1:

src/main/scala/workflow/GatherTransformer.scala
```diff
@@ -4,10 +4,10 @@ import org.apache.spark.rdd.RDD
 
 import scala.reflect.ClassTag
 
-private[workflow] class GatherTransformer[T] extends TransformerNode[Seq[T]] {
-  def transform(dataDependencies: Seq[_], fitDependencies: Seq[TransformerNode[_]]): Seq[T] = dataDependencies.map(_.asInstanceOf[T])
+private[workflow] class GatherTransformer[T] extends TransformerNode {
```
etrain: I don't quite follow here what's going on with the fitDependencies - can you summarize?

src/main/scala/workflow/Node.scala
```diff
@@ -2,6 +2,25 @@ package workflow
 
 import org.apache.spark.rdd.RDD
 
+sealed trait Instruction {
+  def getDependencies: Seq[Int]
+  def mapDependencies(func: Int => Int): Instruction
+}
+
+case class TransformerApplyNode(transformer: Int, inputs: Seq[Int]) extends Instruction {
```
etrain: Can we get some scaladoc on what these classes are?

src/main/scala/workflow/Node.scala
```diff
@@ -9,16 +28,33 @@ sealed trait Node  {
...
-private[workflow] case class DataNode(rdd: RDD[_]) extends Node {
+private[workflow] case class SourceNode(rdd: RDD[_]) extends Node with Instruction {
   override def label: String = "%s[%d]".format(
     Option(rdd.name).map(_ + " ").getOrElse(""), rdd.id)
+  override def getDependencies: Seq[Int] = Seq()
+  override def mapDependencies(func: (Int) => Int): SourceNode = this
 }
+
+/**
+  * A transformer used internally to apply the fit output of an Estimator onto data dependencies.
```
etrain: This is nice!

src/main/scala/workflow/Pipeline.scala
```diff
@@ -47,7 +47,8 @@ trait Pipeline[A, B] {
 
   /**
    * Chains a pipeline onto the end of this one, producing a new pipeline.
-   * @param next the pipeline to chain
+    *
```
etrain: Minor nit - check the spacing here.

etrain:
A lot of this PR is devoted to changing the representation of fit dependencies from `Seq[Seq[Int]]` to `Seq[Option[Int]]`, which feels like a pretty big change in the data structure. Can you explain why this is happening and how the new structure differs from the old one on a conceptual level?

src/test/scala/workflow/DelegatingTransformerSuite.scala
```diff
@@ -1,29 +0,0 @@
-package workflow
```
etrain: Did you mean to delete this file? Is there not some equivalent test we would still want to run?

src/main/scala/workflow/ConcretePipeline.scala
tomerk: Ah probably not, will change

src/main/scala/workflow/WorkflowUtils.scala
```diff
@@ -0,0 +1,124 @@
...
+    current: Int,
+    nodes: Seq[Node],
+    dataDeps: Seq[Seq[Int]],
+    fitDeps: Seq[Option[Int]],
+    nodeIdToInstructionId: Map[Int, Int],
+    instructions: Seq[Instruction]
+  ): (Map[Int, Int], Seq[Instruction]) = {
+    var curIdMap = nodeIdToInstructionId
+    var curInstructions = instructions
+    System.out.println(current)
```
etrain: Probably don't want this.

src/main/scala/workflow/WorkflowUtils.scala
```diff
@@ -0,0 +1,124 @@
...
+    for (dep <- fitDeps(current)) {
+      if (!curIdMap.contains(dep) && dep != Pipeline.SOURCE) {
+        val (newIdMap, newInstructions) = pipelineRecurse(dep, nodes, dataDeps, fitDeps, curIdMap, curInstructions)
+        curIdMap = newIdMap
+        curInstructions = newInstructions
```
etrain: hmm... there's got to be a more functional way to write this. something like `(fitDeps ++ dataDeps).foldLeft(instructions)(thing)` - tracking this mutable state is already making nervous.

src/main/scala/workflow/WorkflowUtils.scala
```diff
@@ -0,0 +1,124 @@
...
+        curIdMap = curIdMap + (current -> curInstructions.length)
+        curInstructions = curInstructions :+ source
+        (curIdMap, curInstructions)
+      }
+
+      case transformer: TransformerNode => {
+        curInstructions = curInstructions :+ transformer
+        val inputs = dataDeps(current).map(curIdMap.apply)
+        curIdMap = curIdMap + (current -> curInstructions.length)
+        curInstructions = curInstructions :+ TransformerApplyNode(curInstructions.length - 1, inputs)
```
etrain: These three blocks are *almost* identical - is there a way this can be tightened?

tomerk:
Sure!
So, before we were using DelegatingTransformers and "fitDependencies" as a way to capture in the DAG that the transformer we intend to apply to the data is the output of an Estimator. Only DelegatingTransformers used these fitDependencies, and there was never any situation in which we would have more than one fit dependency for a DelegatingTransformer.

The `Instruction`s were designed more to match how we've been thinking about these at a higher-level, were the output of an estimator (a transformer) is directly applied to new data. There isn't really a concept captured by these instructions for a situation where a transformer takes data and the outputted transformers of multiple estimators to output new data (which is what multiple fit dependencies would imply currently). (this isn't to suggest that we can't implement instructions later down the line if we decide/need to where a transformer has more than just data dependencies but also dependencies on scala `Any`s)

So, to allow a two-way conversion between the existing Pipeline data structure and a `Seq[Instruction]` : I needed to limit all use of fitDependencies in the existing pipeline data structure to `DelegatingTransformerNode`s, which also entails that there is only ever 0 or 1 fit dependencies for any given node. This is also why I made DelegatingTransformerNode no longer extend TransformerNode.

src/test/scala/workflow/DelegatingTransformerSuite.scala
tomerk: I thought about this earlier. I've stripped this logic out of DelegatingTransformer and put it into the Pipeline apply, so it should be captured by our pipeline chaining and applying unit tests. There isn't really any unit test we can run on an individual DelegatingTransformer object anymore.

src/main/scala/workflow/WorkflowUtils.scala
tomerk: Sure, will do!

etrain:
Makes sense! Thanks for the explanation and the PR. Aside from my general style stuff above this looks pretty good to me.

src/main/scala/workflow/WorkflowUtils.scala
```diff
@@ -0,0 +1,124 @@
...
+  def pipelineToInstructions[A, B](pipeline: Pipeline[A, B]): Seq[Instruction] = {
+    val nodes = pipeline.nodes
+    val dataDeps = pipeline.dataDeps
+    val fitDeps = pipeline.fitDeps
+    val sink = pipeline.sink
+
+    pipelineRecurse(sink, nodes, dataDeps, fitDeps, Map(Pipeline.SOURCE -> Pipeline.SOURCE), Seq())._2
+  }
+
+  def pipelineRecurse(
```
etrain: It would be nice to summarize how this works in a scaladoc.

Something like "Linearizes the pipeline DAG by by walking backwards from the sink in a depth-first manner."

src/main/scala/workflow/WorkflowUtils.scala
```diff
@@ -0,0 +1,124 @@
+package workflow
+
+import scala.collection.mutable.ArrayBuffer
+
+
+object WorkflowUtils {
+  def instructionsToPipeline[A, B](instructions: Seq[Instruction]): Pipeline[A, B] = {
```
etrain: One thing here - is the Pipeline -> Instruction -> Pipeline conversion process fully deterministic and symmetric? That is - if I fed a pipeline through the steps above would I end up with the same exact pipeline? Not that it has to be, but that might be a nice property to be able to rely on..

src/main/scala/workflow/WorkflowUtils.scala
```diff
@@ -0,0 +1,124 @@
...
+            case transformerNode: TransformerNode => {
+              instructionIdToNodeId.put(instruction, nodes.length)
+              nodes.append(transformerNode)
+              dataDeps.append(inputs.map(instructionIdToNodeId.apply))
+              fitDeps.append(None)
+            }
+            case estimatorFitNode: EstimatorFitNode => {
+              instructionIdToNodeId.put(instruction, nodes.length)
+              // TODO: Get more reasonable label...
+              nodes.append(new DelegatingTransformerNode("Estimator fit"))
```
etrain: can we take the .toString() of the parent node and label it `s"Fit[${parent.toString}]"` or something?

src/main/scala/workflow/WorkflowUtils.scala
```diff
@@ -0,0 +1,124 @@
+package workflow
+
+import scala.collection.mutable.ArrayBuffer
+
+
+object WorkflowUtils {
+  def instructionsToPipeline[A, B](instructions: Seq[Instruction]): Pipeline[A, B] = {
+    val nodes = new ArrayBuffer[Node]()
+    val dataDeps = new ArrayBuffer[Seq[Int]]()
+    val fitDeps = new ArrayBuffer[Option[Int]]()
```
etrain: would this be easier to write with a recursive/functional definition rather than one where we build up the state mutably? It doesn't matter becuase all these datastructures are internal, but i'm thinking about how to tighten up/reduce code complexity.

src/main/scala/workflow/WorkflowUtils.scala
tomerk: It is not exactly fully deterministic and symmetric, but I believe if you do:
`x = Pipeline -> Instruction -> Pipeline`
Then `x -> Instruction -> Pipeline == x`
Although I wouldn't want to make any guarantees on that.

src/main/scala/workflow/WorkflowUtils.scala
tomerk: I could replace this with a `foldLeft` and a case class w/ a merge method. If you prefer that, it might tighten the code up somewhat (up to 9 lines in this method or so, then it would add maybe 3 - 5 lines elsewhere), and it would remove the mutability.

etrain:
Could you add some unit tests for the pipeline -> instructions stuff and back to make sure that we end up with valid transformations? Should handle the edge cases as well as a few common cases.

src/main/scala/workflow/WorkflowUtils.scala
etrain: I think that would be great. My gut feeling is that this opens the door for fewer bugs down the road and increases maintainability/extensibility.

etrain:
Other than the test I just asked for and the comment on the Instructions -> Pipeline transformation style, this looks pretty good to me.

tomerk:
@etrain how's it look now?

etrain:
Ok - this LGTM. I'm going to go ahead and merge this and we can move on to the next PR. Thanks @tomerk !

PR #222, opened 2016-02-01
Added a node-level optimization rule

tomerk:
@etrain @shivaram

### DIFF
--- a/src/main/scala/workflow/Node.scala
+++ b/src/main/scala/workflow/Node.scala
@@ -5,7 +5,15 @@ import org.apache.spark.rdd.RDD
 sealed trait Instruction {
   def getDependencies: Seq[Int]
   def mapDependencies(func: Int => Int): Instruction
+  def execute(deps: Seq[InstructionOutput]): InstructionOutput
 }
+
+sealed trait InstructionOutput
+abstract class DataOutput extends InstructionOutput
+case class RDDOutput(rdd: RDD[_]) extends DataOutput
+case class SingleDataOutput(data: Any) extends DataOutput
+case class EstimatorOutput(estimatorNode: EstimatorNode) extends InstructionOutput
+case class TransformerOutput(transformerNode: TransformerNode) extends InstructionOutput
 
 /**
  * An Instruction that represents the application of a transformer to data inputs.
@@ -17,6 +25,18 @@ case class TransformerApplyNode(transformer: Int, inputs: Seq[Int]) extends Inst
   override def getDependencies: Seq[Int] = Seq(transformer) ++ inputs
   override def mapDependencies(func: (Int) => Int): TransformerApplyNode = {
     TransformerApplyNode(func(transformer), inputs.map(func))
+  }
+
+  override def execute(deps: Seq[InstructionOutput]): DataOutput = {
+    val transformer = deps.collectFirst {
+      case TransformerOutput(t) => t
+    }.get
+
+    val dataInputs = deps.tail.collect {
+      case RDDOutput(rdd) => rdd
+    }
+
+    RDDOutput(transformer.transformRDD(dataInputs))
   }
 }
 
@@ -31,6 +51,18 @@ case class EstimatorFitNode(est: Int, inputs: Seq[Int]) extends Instruction {
   override def mapDependencies(func: (Int) => Int): EstimatorFitNode = {
     EstimatorFitNode(func(est), inputs.map(func))
   }
+
+  override def execute(deps: Seq[InstructionOutput]): TransformerOutput = {
+    val estimator = deps.collectFirst {
+      case EstimatorOutput(e) => e
+    }.get
+
+    val dataInputs = deps.tail.collect {
+      case RDDOutput(rdd) => rdd
+    }
+
+    TransformerOutput(estimator.fitRDDs(dataInputs))
+  }
 }
 
 sealed trait Node  {
@@ -44,6 +76,7 @@ private[workflow] abstract class EstimatorNode extends Node with Serializable wi
   private[workflow] def fitRDDs(dependencies: Seq[RDD[_]]): TransformerNode
   override def getDependencies: Seq[Int] = Seq()
   override def mapDependencies(func: (Int) => Int): EstimatorNode = this
+  override def execute(deps: Seq[InstructionOutput]): EstimatorOutput = EstimatorOutput(this)
 }
 
 private[workflow] abstract class TransformerNode extends Node with Serializable with Instruction {
@@ -52,6 +85,7 @@ private[workflow] abstract class TransformerNode extends Node with Serializable 
 
   override def getDependencies: Seq[Int] = Seq()
   override def mapDependencies(func: (Int) => Int): TransformerNode = this
+  override def execute(deps: Seq[InstructionOutput]): TransformerOutput = TransformerOutput(this)
 }
 
 private[workflow] case class SourceNode(rdd: RDD[_]) extends Node with Instruction {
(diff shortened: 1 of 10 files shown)


src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -0,0 +1,94 @@
...
+          case _: OptimizableLabelEstimator[_, _, _] => WorkflowUtils.getParents(index, instructions) + index
+          case _ => Set[Int]()
+        }
+
+      case _ => Set[Int]()
+    }.reduce(_ union _) -- WorkflowUtils.getChildren(Pipeline.SOURCE, instructions) - Pipeline.SOURCE
+
+    // Execute the minimal amount necessary of the pipeline on sampled nodes, and optimize the optimizable nodes
+    val optimizedInstructions = instructions.toArray
+    val registers = new Array[Any](instructions.length)
```
etrain: Sorry - I'm not following. What are the registers?

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -0,0 +1,94 @@
...
+
+          case transformer: TransformerNode =>
+            registers(index) = transformer
+
+          case estimator: EstimatorNode =>
+            registers(index) = estimator
+
+          case TransformerApplyNode(tIndex, inputIndices) =>
+            val numPerPartition = numPerPartitionPerNode(inputIndices.head)
+            val inputs = inputIndices.map(registers(_).asInstanceOf[RDD[_]].cache())
```
etrain: all these `.asInstanceOf[]`are making me a bit uneasy - what is all this stemming from? I realize the registers are an `Array[Any]` but what are they representing?

etrain:
The design/interfaces all look pretty reasonable to me, since they're pretty much what we discussed in our design meetings. However, I'm having a really hard time following the logic inside the node-level optimizer probably because the types are strange.

src/main/scala/workflow/NodeOptimizationRule.scala
tomerk: Think of this as an interpreter for static single assignment form bytecode (like we've discussed in the past). `registers` is the structure that stores the variable assignments. The output value of each instruction is assigned to the variable `i`, where `i` is the index of the instruction. Would making it a `Map[Int, Any]` instead of an `Array[Any]`, or renaming it to something else make it easier to follow?

src/main/scala/workflow/NodeOptimizationRule.scala
tomerk: Does this make more sense given the above explanation of what the registers are? It's basically just typecasting the stored variables to the type that the instruction expects.

tomerk:
@etrain How do you feel about how it looks like this?

src/main/scala/workflow/Node.scala
```diff
@@ -5,8 +5,16 @@ import org.apache.spark.rdd.RDD
...
 
+sealed trait InstructionOutput
+abstract class DataOutput extends InstructionOutput
+case class RDDOutput(rdd: RDD[_]) extends DataOutput
+case class SingleDataOutput(data: Any) extends DataOutput
+case class EstimatorOutput(estimatorNode: EstimatorNode) extends InstructionOutput
+case class TransformerOutput(transformerNode: TransformerNode) extends InstructionOutput
+
 /**
  * An Instruction that represents the application of a transformer to data inputs.
```
etrain: Is this correct? Is an instruction always a transformer?

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -0,0 +1,96 @@
...
+              case RDDOutput(rdd) => rdd.cache()
+            }
+            inputs.foreach(_.count())
+
+            // Optimize the estimator if possible
+            val estimator = registers(estIndex) match {
+              case EstimatorOutput(oe: OptimizableEstimator[a, b]) =>
+                oe.optimize(inputs.head.asInstanceOf[RDD[a]], numPerPartition.get)
+              case EstimatorOutput(oe: OptimizableLabelEstimator[a, b, l]) =>
+                oe.optimize(inputs.head.asInstanceOf[RDD[a]], inputs(1).asInstanceOf[RDD[l]], numPerPartition.get)
```
etrain: Style nit - is there a reason not to use `(0)` in places of `.head` here since you're using `(1)` in the same line?

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -0,0 +1,96 @@
...
+
+    // Execute the minimal amount necessary of the pipeline on sampled nodes, and optimize the optimizable nodes
+    val optimizedInstructions = instructions.toArray
+    val registers = new Array[InstructionOutput](instructions.length)
+    val numPerPartitionPerNode = Array.fill[Option[Map[Int, Int]]](instructions.length)(None)
+
+    for ((instruction, index) <- instructions.zipWithIndex) {
+      if (instructionsToExecute.contains(index)) {
+        instruction match {
+          case SourceNode(rdd) =>
```
etrain: style nit - can we put braces around the things after the case statements - it is kind of hard to see which expressions are logically grouped here (this is probably my C days talking but i get scared when things aren't protected by curly braces that maybe need to be.

src/test/scala/workflow/NodeOptimizationRuleSuite.scala
```diff
@@ -0,0 +1,161 @@
...
+
+    val nodeOptimizedPipeline = new NodeOptimizationRule(0.01).apply(pipeline)
+    val outputState = nodeOptimizedPipeline.apply(State(), optimizer = None)
+
+    assert(outputState === State(None, Some(false), Some(true), Some(true)))
+  }
+}
+
+object NodeOptimizationRuleSuite {
+  case class State(choice: Option[Boolean] = None, transformerChoice: Option[Boolean] = None, estimatorChoice: Option[Boolean] = None, labelEstimatorChoice: Option[Boolean] = None)
```
etrain: Looks like a pretty long line.

src/test/scala/workflow/WorkflowUtilsSuite.scala
```diff
@@ -57,7 +57,7 @@ class WorkflowUtilsSuite extends FunSuite with LocalSparkContext with Logging {
       (plusDataPlusLabelEstimator, trainData, labelData)
 
     val initialPipeline = new Optimizer {
```
etrain: Gotta say - the ability to dynamically construct different optimizers with `::` is awesome.

src/test/scala/workflow/NodeOptimizationRuleSuite.scala
```diff
@@ -0,0 +1,161 @@
...
+import NodeOptimizationRuleSuite._
+
+import org.apache.spark.SparkContext
+import org.apache.spark.rdd.RDD
+import org.scalatest.FunSuite
+import pipelines.{LocalSparkContext, Logging}
+
+import scala.util.Random
+
+class NodeOptimizationRuleSuite extends FunSuite with LocalSparkContext with Logging {
```
etrain: These tests are great!

etrain:
This is looking really good. A few style nits but otherwise I'm happy to merge.

src/main/scala/workflow/Node.scala
tomerk: Sorry I guess this comment isn't clear enough. I just mean TransformerApplyNode is an instruction that represents the application of a transformer to data inputs (and in turn outputs data).

src/main/scala/workflow/NodeOptimizationRule.scala
tomerk: nope no reason besides IntelliJ highlighting it yellow and trying to autocorrect it to head, I can make it `inputs(0)`

tomerk:
Okay @etrain I've made the style changes you asked for.
I assume you're happier w/ the clarity of what the NodeOptimizationRule is doing with the `Seq[Instruction]` now?

etrain:
Looks much better in both regards! Merging this and we can decide if we need to patch what a Node level optimization rule actually means later.

PR #223, opened 2016-02-01
Unified sparse and dense for some of the solvers

tomerk:
@etrain @shivaram


src/main/scala/nodes/learning/BlockLinearMapper.scala
```diff
@@ -18,24 +18,24 @@ import workflow.{Transformer, LabelEstimator}
  * @param bOpt optional intercept term to be added
  * @param featureScalersOpt optional seq of transformers to be applied before transformation
  */
-class BlockLinearMapper(
+class BlockLinearMapper[T <: Vector[Double]](
     val xs: Seq[DenseMatrix[Double]],
     val blockSize: Int,
     val bOpt: Option[DenseVector[Double]] = None,
     val featureScalersOpt: Option[Seq[Transformer[DenseVector[Double], DenseVector[Double]]]] = None)
```
etrain: Does it make sense for this to take `Option[Seq[Transformer[DenseVector[Double], DenseVector[Double]]]]]` or `O[S[T[Vector[Double], Vector[Double]]]]`

etrain: Also - that's a *very* complicated type signature.

tomerk: We could change it to `Option[Seq[StandardScalerModel]]` I think

src/main/scala/nodes/learning/LBFGS.scala
```diff
@@ -130,22 +130,27 @@ object LBFGSwithL2 extends Logging {
...
+  def fit(data: RDD[T], labels: RDD[DenseVector[Double]]): LinearMapper[T] = {
+    val denseData = data.map {
+      case denseVec: DenseVector[Double] => denseVec
+      case vec => vec.toDenseVector
+    }
 
-  def fit(data: RDD[DenseVector[Double]], labels: RDD[DenseVector[Double]]): LinearMapper = {
     if (fitIntercept) {
-      val featureScaler = new StandardScaler(normalizeStdDev = false).fit(data)
+      val featureScaler = new StandardScaler(normalizeStdDev = false).fit(denseData)
```
etrain: Hmm... why should `DenseLBFGSWithL2` accept sparse data at all? Is there something we *can't* do if we leave the interface as Dense only?

src/main/scala/nodes/learning/LBFGS.scala
```diff
@@ -177,19 +182,24 @@ class DenseLBFGSwithL2(
  * @param regParam L2 regularization
  * @param numIterations max number of iterations to run
  */
-class SparseLBFGSwithL2(
+class SparseLBFGSwithL2[T <: Vector[Double]](
```
etrain: Same thing here, but in reverse.

etrain:
Hmm... code-wise this looks OK. But I guess the big thing is whether we think this is the right interface for solvers - e.g. should they really take a single type and then dispatch automatically?

Also - if we do it the way its currently set up why would we have separate classes for Sparse/Dense LBFGS?

@shivaram what do you think?

src/main/scala/nodes/learning/LBFGS.scala
tomerk: So, some background on why I did this sparse and dense unification stuff:
Because an optimizable estimator can only output a single estimator (as opposed to a transformer chained to an estimator), the optimizable linear solver I wrote can only handle both sparse and dense if all of the estimator choices can handle both sparse and dense. This means both `DenseLBFGSWithL2` and `SparseLBFGSWithL2` need to be able to take either sparse or dense data as input.

tomerk:
Wrote this above but will put here:

Some background on why I did this sparse and dense unification stuff:
An optimizable node can only output a single optimized node (as opposed to a transformer chained to an estimator). So, the optimizable linear solver can only handle both sparse and dense if all of the estimator choices can handle both sparse and dense. This means both DenseLBFGSWithL2 and SparseLBFGSWithL2 need to be able to take either sparse or dense data as input.

As for why we would have separate classes for Sparse/Dense LBFGS: The intercept fitting strategies for the two are different, but I could unify them if you guys like.

src/main/scala/nodes/learning/BlockLinearMapper.scala
shivaram: Yeah lets make it `StandardScalerModel` (or invent a new class if we need)

src/main/scala/nodes/learning/LBFGS.scala
```diff
@@ -130,22 +130,27 @@ object LBFGSwithL2 extends Logging {
...
     val convergenceTol: Double = 1e-4,
     val numIterations: Int = 100,
     val regParam: Double = 0.0)
-  extends LabelEstimator[DenseVector[Double], DenseVector[Double], DenseVector[Double]] {
+  extends LabelEstimator[T, DenseVector[Double], DenseVector[Double]] {
+
+  def fit(data: RDD[T], labels: RDD[DenseVector[Double]]): LinearMapper[T] = {
+    val denseData = data.map {
+      case denseVec: DenseVector[Double] => denseVec
+      case vec => vec.toDenseVector
```
shivaram: Rather than transform it, shouldn't we throw some error here ?

src/main/scala/nodes/learning/LinearMapper.scala
```diff
@@ -39,10 +43,15 @@ case class LinearMapper(
...
    */
-  override def apply(in: RDD[DenseVector[Double]]): RDD[DenseVector[Double]] = {
+  override def apply(in: RDD[T]): RDD[DenseVector[Double]] = {
     val modelBroadcast = in.context.broadcast(x)
     val bBroadcast = in.context.broadcast(bOpt)
-    val inScaled = featureScaler.map(_.apply(in)).getOrElse(in)
+
+    val inScaled = featureScaler.map(_.apply(in.map {
+      case dense: DenseVector[Double] => dense
+      case notDense => notDense.toDenseVector
```
shivaram: `case _ =>` ? instead of `notDense` that is

src/main/scala/nodes/learning/SparseLinearMapper.scala
```diff
@@ -1,50 +0,0 @@
...
-import org.apache.spark.rdd.RDD
-import workflow.Transformer
-
-/**
- * Computes A * x + b i.e. a linear map of data using a trained model.
- *
- * @param x trained model
- * @param bOpt optional intercept to add
- */
-case class SparseLinearMapper(
```
shivaram: Yay  ! Deleting code is the best thing ever

shivaram:
I'm a little confused about which way to see this 

a. If the solvers take in a single type `Vector` and then dispatch, the nice thing is that it means we separate out the kind of optimization (LBFS vs. co-ordinate descent etc.) from the Sparse / Dense stuff

b. However if we are going to write a single optimizer which picks a solver class then having separate nodes for SparseLBFGS, DenseLBFGS is useful as the optimizer can pick a class.

src/main/scala/nodes/learning/LBFGS.scala
tomerk: My thinking was that if we type this estimator to work on any data that extends `Vector`, then it should be able to work on SparseVectors, converting to Dense if necessary. There are also cases where if the sparse vector has a high density, it's better to just make it fully dense anyway rather than trying to use sparse operations.

tomerk:
We're limited to going with option a, even with a single optimizer that picks a solver class. The node-level optimizer requires all of the optimized choices to have the exact same type signature.

etrain:
So if I understand correctly, the reason this is hard is because the contract for the `OptimizableNode` interface is that it returns an `Estimator`. What if it could return a `Pipeline`? Is it not the case that an `Estimator` is a pipeline? 

If so, the logic would be simply to merge the `Pipeline` in in the right place. 

The reason I'm making a big deal about this is that I can easily imagine situations where we decide to replace a single node with a composition of nodes in other cases - e.g. if someone puts an `RBFEstimator` we might just substitute with a `CosineRandomFeatures andThen LeastSquaresEstimator` or something like that. Having the full DAG is useful at later steps, etc.

tomerk:
That makes sense, but an Estimator by itself isn't a Pipeline. An Estimator with training data attached to it is a Pipeline. 

It may be possible to have the optimizableEstimator return `(trainData: RDD[_]) => Pipeline`, and then somehow merge in the pipeline once we have the real trainData (not just a sample of it) available as we're executing the Pipeline. (I don't have a clear picture of how the code for this would look in my head, we would need to whiteboard it out).

Alternatively, it may be possible to write a utility to chain a transformer to an Estimator w/o data attached to output a new estimator.

etrain:
I'm in favor of the general solution rather than the special case that solves the problem we have right now. Happy to whiteboard this tomorrow.

tomerk:
Closing because no longer needed

PR #224, opened 2016-02-06
Auto caching rule

tomerk:
Added auto-caching rules w/ aggressive naive, and greedy profiling strategies
@etrain 
@shivaram

### DIFF
--- a/src/main/scala/workflow/AutoCacheRule.scala
+++ b/src/main/scala/workflow/AutoCacheRule.scala
@@ -11,9 +11,7 @@ case class Profile(ns: Long, mem: Long) {
 
 case class SampleProfile(scale: Long, profile: Profile)
 
-class AutoCacheRule(
-  cachingMode: CachingStrategy
-) extends Rule with Logging {
+class AutoCacheRule(cachingMode: CachingStrategy) extends Rule with Logging {
 
   /**
    * Get the node weights: estimates for how many passes an instruction will make over its input dependencies
--- a/src/main/scala/workflow/AutoCacheRule.scala
+++ b/src/main/scala/workflow/AutoCacheRule.scala
@@ -1,0 +1,257 @@
+package workflow
+
+import breeze.linalg.{max, DenseVector, DenseMatrix}
+import nodes.util.Cacher
+import org.apache.spark.rdd.RDD
+import pipelines.Logging
+
+case class Profile(ns: Long, mem: Long) {
+  def +(p: Profile) = Profile(this.ns + p.ns, this.mem + p.mem)
+}
+
+case class SampleProfile(scale: Long, profile: Profile)
+
+class AutoCacheRule(
+                     profileScales: Seq[Long] = Seq(500, 1000),
+                     numProfileTrials: Int = 2
+                   ) extends Rule with Logging {
+
+  /**
+   * Get the node weights: estimates for how many passes an instruction will make over its input dependencies
+   */
+  def getNodeWeights(instructions: Seq[Instruction]): Map[Int, Int] = {
+    instructions.zipWithIndex.map {
+      case (TransformerApplyNode(transformer, _), i) => instructions(transformer) match {
+        case node: WeightedNode => (i, node.weight)
+        case _ => (i, 1)
+      }
+      case (EstimatorFitNode(estimator, _), i) => instructions(estimator) match {
+        case node: WeightedNode => (i, node.weight)
+        case _ => (i, 1)
+      }
+      case (_, i) => (i, 1)
+    }.toMap
+  }
+
+  /**
+   * Get a map representing the immediate children for each instruction
+   * Note: This doesn't capture how many times each child depended on the instruction
+   */
+  def getImmediateChildrenByInstruction(instructions: Seq[Instruction]): Map[Int, Set[Int]] = {
+    instructions.indices.map(i => (i, WorkflowUtils.getChildren(i, instructions))).toMap
+  }
+
+  /**
+   * Get an estimate for how many times each instruction will be executed, assuming
+   * the given set of instructions have their outputs cached
+   */
+  def getRuns(instructions: Seq[Instruction], cache: Set[Int], nodeWeights: Map[Int, Int]): Map[Int, Int] = {
+    val immediateChildrenByInstruction = getImmediateChildrenByInstruction(instructions)
+
+    instructions.indices.foldRight(Map[Int, Int]()) { case (i, runsByIndex) =>
+      if (immediateChildrenByInstruction(i).isEmpty) {
+        runsByIndex + (i -> 1)
+      }
+      else {
+        val runs = immediateChildrenByInstruction(i).map(j => if (cache.contains(j)) {
+          nodeWeights(j)
+        } else {
+          nodeWeights(j) * runsByIndex(j)
+        }).sum
+        runsByIndex + (i -> runs)
+      }
+    }
+  }
+
+  def generalizeProfiles(newScale: Long, sampleProfiles: Seq[SampleProfile]): Profile = {
+    def getModel(inp: Iterable[(Long, String, Long)]): Double => Double = {
+      val observations = inp.toArray
+
+      //Pack a data matrix with observations
+      val X = DenseMatrix.ones[Double](observations.length, 2)
+      observations.zipWithIndex.foreach(o => X(o._2, 0) = o._1._1.toDouble)
+      val y = DenseVector(observations.map(_._3.toDouble))
+      val model = max(X \ y, 0.0)
+
+      //A function to apply the model.
+      def res(x: Double): Double = DenseVector(x, 1.0).t * model
+
+      res
+    }
+
+    val samples = sampleProfiles.flatMap { case SampleProfile(scale, value) =>
+      Array(
+        (scale, "memory", value.mem),
+        (scale, "time", value.ns)
+      )}.groupBy(a => a._2)
+
+    val models = samples.mapValues(getModel)
+
+    Profile(models("time").apply(newScale).toLong, models("memory").apply(newScale).toLong)
+  }
+
+  def profileInstructions(
+                           instructions: Seq[Instruction],
+                           scales: Seq[Long],
+                           numTrials: Int
+                         ): Map[Int, Profile] = {
+    val instructionsToProfile = instructions.indices.toSet -- WorkflowUtils.getChildren(Pipeline.SOURCE, instructions)
+
+    val registers = scala.collection.mutable.Map[Int, Any]()
+    val numPerPartitionPerNode = scala.collection.mutable.Map[Int, Map[Int, Int]]()
+    val profiles = scala.collection.mutable.Map[Int, Profile]()
+
+    val sortedScales = scales.sorted
+    for ((instruction, i) <- instructions.zipWithIndex if instructionsToProfile.contains(i)) {
+      instruction match {
+        case SourceNode(rdd) =>
+          val npp = WorkflowUtils.numPerPartition(rdd)
+          numPerPartitionPerNode(i) = npp
+
+          val totalCount = npp.values.map(_.toLong).sum
+
+          val sampleProfiles = for (
+            (scale, scaleIndex) <- sortedScales.zipWithIndex;
+            trial <- 1 to numTrials
+          ) yield {
+            // Calculate the necessary number of items per partition to maintain the same partition distribution,
+            // while only having scale items instead of totalCount items.
+            // Can't use mapValues because that isn't serializable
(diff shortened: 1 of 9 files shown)


src/main/scala/workflow/AutoCacheRule.scala
```diff
@@ -0,0 +1,357 @@
...
+import workflow.AutoCacheRule.{GreedyCache, NaiveCache, CachingStrategy}
+
+case class Profile(ns: Long, mem: Long) {
+  def +(p: Profile) = Profile(this.ns + p.ns, this.mem + p.mem)
+}
+
+case class SampleProfile(scale: Long, profile: Profile)
+
+class AutoCacheRule(
+  cachingMode: CachingStrategy
```
shivaram: will fit in previous line

etrain:
At a high level this looks pretty good. I'll provide some more detailed comments on the code shortly.

A couple things I'm missing:
1) How do you think a user would actually use the autocaching rule? Can you add it as an Optimizer option or something? 
2) In that same vein - it might be nice to actually test this on (say) the VOC pipeline and ensure that a) profiling/caching is reasonably fast and b) that the outcomes are correct. The bulk of stuff needed for that is in my dag-estimation branch.

src/main/scala/workflow/AutoCacheRule.scala
```diff
@@ -0,0 +1,355 @@
...
+object AutoCacheRule {
+  sealed trait CachingStrategy
+  /**
+   * NaiveCache is the optimal caching strategy assuming infinite memory and zero caching overhead
+   * (cache at every node with > 1 direct output dependency edge during training)
+   */
+  case object NaiveCache extends CachingStrategy
+  case class GreedyCache(
+    maxMem: Long,
+    profileScales: Seq[Long] = Seq(500, 1000),
```
etrain: Hmm... does it make sense to have a fixed constant here as default? Can we figure out a way ot make it more data dependent?

src/main/scala/workflow/AutoCacheRule.scala
```diff
@@ -0,0 +1,355 @@
...
+          nodeWeights(j)
+        } else {
+          nodeWeights(j) * runsByIndex(j)
+        }).sum
+        runsByIndex + (i -> runs)
+      }
+    }
+  }
+
+  def generalizeProfiles(newScale: Long, sampleProfiles: Seq[SampleProfile]): Profile = {
```
etrain: This method is complicated enough that it probably needs some scaladoc - the important assumption here is that we're dealing with a Seq[SampleProfile] for a given node, right?

src/main/scala/workflow/AutoCacheRule.scala
```diff
@@ -0,0 +1,355 @@
...
+          oldToNewIndexMap.map {
+            case (key, value) => if (key >= i) (key, value + 2) else (key, value)
+          })
+
+      case ((newPipe, oldToNewIndexMap), i) =>
+        (newPipe :+ pipe(i).mapDependencies(oldToNewIndexMap), oldToNewIndexMap)
+    }._1
+  }
+
+  def naiveCache(instructions: Seq[Instruction]): Seq[Instruction] = {
```
etrain: Can we call this something other than `naive` since that's not very descriptive anyway? Something that conveys that we're caching everything that is being reused and with an LRU replacement policy.

src/main/scala/workflow/AutoCacheRule.scala
```diff
@@ -0,0 +1,355 @@
...
+  }
+
+  def generalizeProfiles(newScale: Long, sampleProfiles: Seq[SampleProfile]): Profile = {
+    def getModel(inp: Iterable[(Long, String, Long)]): Double => Double = {
+      val observations = inp.toArray
+
+      //Pack a data matrix with observations
+      val X = DenseMatrix.ones[Double](observations.length, 2)
+      observations.zipWithIndex.foreach(o => X(o._2, 0) = o._1._1.toDouble)
+      val y = DenseVector(observations.map(_._3.toDouble))
```
shivaram: On the above note, it'll be good to add a note on what we are solving for in the line below.

src/main/scala/workflow/AutoCacheRule.scala
tomerk: Well I'm pretty sure the whole point of setting an explicit integer constant instead of a data ratio was specifically to avoid dependence on the amount of data.

I'm not sure what the right way to set this number based off data properties would be. We could just make it not default to anything and the user would need to choose the scales.

src/main/scala/workflow/AutoCacheRule.scala
tomerk: Yes, that's correct. I'll add some scaladoc.

src/main/scala/workflow/AutoCacheRule.scala
etrain: So - let's consider the VOC case - at 5000 examples the time spent profiling would be 30% of the total runtime in the best case, which seems awfully high. I *think* the way I had it before was as a function of the number of partitions (2 or 4 examples per partition, for example). I don't think we want users to have to explicitly choose this parameter (though it should be configurable) - instead we should pick something that we think would work and not slow things down too much in most cases. We have a few workloads that we can think about in this context, right? E.G. CIFAR is going to be 50k tiny images spread in nummachines*numcores partitions in the usual case, Amazon is 65m reviews spread across a similar number of partitions, TIMIT is 2m examples, and Solar Flares is some smaller number of images. How would each of these situations behave?

src/main/scala/workflow/AutoCacheRule.scala
```diff
@@ -0,0 +1,355 @@
...
+  ): Map[Int, Profile] = {
+    val instructionsToProfile = instructions.indices.toSet -- WorkflowUtils.getChildren(Pipeline.SOURCE, instructions)
+
+    val registers = new Array[InstructionOutput](instructions.length)
+    val numPerPartitionPerNode = scala.collection.mutable.Map[Int, Map[Int, Int]]()
+    val profiles = scala.collection.mutable.Map[Int, Profile]()
+
+    val sortedScales = scales.sorted
+    for ((instruction, i) <- instructions.zipWithIndex if instructionsToProfile.contains(i)) {
+      instruction match {
```
etrain: I feel like I'm missing something here - is there a reason we don't profile Estimators here? Or do we profile them but I'm not seeing where?

tomerk: We don't profile them for two reasons:
1. At the moment the output of an estimator (transformer) is always cached no matter what, so there's no need to profile them.
2. I doubt the profile generalization work correctly anyway (we would need Planck for that)

src/main/scala/workflow/AutoCacheRule.scala
tomerk: Sure, I can rename it. `MinimalLRUCache` or something like that.

src/main/scala/workflow/AutoCacheRule.scala
tomerk: I honestly don't have any sense of what the right thing to do would be in each of those situations. What's the minimal number of data points needed for each to provide reasonable / useful profiling info?

src/main/scala/workflow/AutoCacheRule.scala
etrain: Hmm.. I feel like the property that the runtime estimate semi-accurately reflects reality is a good one. For example, during the initial dag-estimation work, I verified that the estimates were within 20% of reality in terms of total execution time for the whole pipeline.

But, I agree that sampling fewer things leads to more efficient optimization. However, from a research perspective I have to believe that having profiles for every node in the pipeline would enable a much larger number of algorithms - seems like we should at least have a code path that enables this (even if via an override of `instructionsToProfile` or something).

src/main/scala/workflow/AutoCacheRule.scala
etrain: So on the memory side of things this is super easy - in general, we see almost perfect scaling with `n` in terms of memory usage. 

On the time side of things, it is a little tricker. There's some constant overheads from firing off a spark job, but then (at least for transformers) basically linear time with the number of data points. 

There are obviously some details about transferring data to/from the driver, etc, but this is meant to be a rough estimate of execution time. I think that a small handful of points/partition is "enough" to get a good estimate of total runtime, but this feels like something we should be able to test.

src/main/scala/workflow/AutoCacheRule.scala
etrain: `MinimalLRUCache` seems like a misnomer to me, since it implies miminalism. But `AggressiveCache` seems ok.

src/main/scala/workflow/AutoCacheRule.scala
tomerk: Hmm, if you think that the linear profile generalization would be sufficiently accurate then I can go ahead and add some code to enable this.

src/main/scala/workflow/AutoCacheRule.scala
tomerk: Which of the workloads did you already test and which do we still need to?

src/main/scala/workflow/AutoCacheRule.scala
tomerk: done.

etrain:
Given that auto-caching adds profiling overheads that we don't understand fully, can we make it so that it is *not* a default optimization? It would be great to then add a flag somewhere that allows it to be turned on easily, so that users can find out if it helps their pipeline.

src/main/scala/workflow/AutoCacheRule.scala
```diff
@@ -0,0 +1,531 @@
...
+        } else {
+          nodeWeights(j) * runsByIndex(j)
+        }).sum
+        runsByIndex + (i -> runs)
+      }
+    }
+  }
+
+  /**
+   * Get the initial set of what instructions will have their results effectively cached
```
etrain: Can you elaborate on what this function is supposed to do? I'm not sure I follow, especially because if this is doing what I think its doing, I would expect `TransformerNode` to return `false` or `instructions(t).isInstanceOf[Cacher[_]]`.

src/main/scala/workflow/AutoCacheRule.scala
```diff
@@ -0,0 +1,531 @@
...
+    maxMem: Option[Long]
+  ): Seq[Instruction] = {
+    val nodeWeights = getNodeWeights(instructions)
+
+    var cached = initCacheSet(instructions)
+    val memBudget = maxMem.getOrElse(instructions.collectFirst {
+      case SourceNode(rdd) =>
+        val sc = rdd.sparkContext
+        // Calculate memory budget, ignoring driver unless in local mode
+        val workers = if (sc.getExecutorStorageStatus.size == 1) {
```
etrain: Hmm... I'm not sure if this is a canonical way to check this. What if I'm running in `local[4]`?

etrain:
Alright - I have given this a pretty thorough read through, and I think it is basically correct, modulo the few minor issues I pointed out. Will give it a second read for style, but if you could address the two things I pointed out (both of them may be fine, I just wanted to double check) then I think we are good to merge this one.

src/main/scala/workflow/AutoCacheRule.scala
tomerk: With `local[4]`, getExecutorStorageStatus returns a single executor status, the driver's. This would go down the first code path and return the correct memory budget.

src/main/scala/workflow/AutoCacheRule.scala
tomerk: So in a sense, the Transformers themselves are ALWAYS cached because they always take up driver memory, it's the output of a transformer being applied to data that isn't necessarily cached. When you see a `TransformerNode` as an instruction, it just signals to store that transformer in the registers, not to apply it to anything. Likewise for `EstimatorNodes` and estimators.

tomerk:
@etrain So do you want me to make a System flag `-Dkeystoneml.autocache`, or just remove auto-caching from the default optimizer and provide a separate optional optimizer that does auto-caching?

etrain:
I think resisting system flags as long as possible is a good idea. The
latter proposal is good!

src/main/scala/workflow/AutoCacheRule.scala
etrain: Right - Ok, makes sense now. Basically this shows I lack a good mental
model for the registers/instructions machine - gets better with every PR
reviewed, though :)

tomerk:
Okay @etrain, done

etrain:
Awesome work. Thanks @tomerk!

Issue #183, opened 2015-12-08
Misc Changes from keystone-vw-experiments


tomerk:
Bring in bug fixes to:

- [x] LinearMapper cost function
- [x] CommonSparseFeatures
- [X] MatrixUtils

(etc. whatever I find hasn't made it into master yet)

etrain:
@tomerk - just want to ping you on this so that we don't forget it. not sure what needs to be done with CommonSparseFeatures or LinearMapper (maybe the latter is good now?)

shivaram:
I think CommonSparseFeatures had some fix to make it more scalable ? Did we already port that back ?

tomerk:
Well the thing about CommonSparseFeatures is that we've made it more scalable for large numbers of features, but it's probably slower for small numbers of features. I could make two different nodes, or just use the implementation that's more scalable for large numbers of features.

tomerk:
And then the LinearMapper thing is just a utility method to calculate loss, I'll make a PR.

shivaram:
I'd say we should pick the more scalable one unless its much slower with smaller number of features. Do you have some numbers for what kind of slow down we are talking about ? Anyways its worth opening a PR for this and discussing it there

etrain:
+1 for @shivaram's suggestion - if you could sketch out the two alternatives in the comments for the PR we can decide there.

tomerk:
@shivaram: Nope I don't have timing numbers. The main difference is we fully sort each partition instead of doing an optimal get-top-k, so the per-partition time would be n log n instead of n log k.
(where n is number of features in that partition, and k is the total number of desired features)

tomerk:
These have all gotten merged in at different points in time

PR #225, opened 2016-02-08
made commonsparsefeatures more scalable...

tomerk:
at large numbers of machines and high feature counts.

See issue #183

### DIFF
--- a/src/main/scala/nodes/util/CommonSparseFeatures.scala
+++ b/src/main/scala/nodes/util/CommonSparseFeatures.scala
@@ -4,6 +4,7 @@ import breeze.linalg.SparseVector
 import org.apache.spark.rdd.RDD
 import workflow.Estimator
 
+import scala.collection.JavaConversions._
 import scala.reflect.ClassTag
 
 /**
@@ -27,7 +28,10 @@ case class CommonSparseFeatures[T : ClassTag](numFeatures: Int) extends Estimato
     }
   }
 
-  def merge(a: Array[(T, (Int, Long))], b: Array[(T, (Int, Long))]): Array[(T, (Int, Long))] = (a ++ b).sorted(ordering.reverse).take(numFeatures)
+  /** This method merges two seqs and keeps the top numFeatures */
+  def merge(a: Seq[(T, (Int, Long))], b: Seq[(T, (Int, Long))]): Seq[(T, (Int, Long))] = {
+    (a ++ b).sorted(ordering.reverse).take(numFeatures)
+  }
 
   override def fit(data: RDD[Seq[(T, Double)]]): SparseFeatureVectorizer[T] = {
     val featureOccurrences = data.flatMap(identity).zipWithUniqueId().map(x => (x._1._1, (1, x._2)))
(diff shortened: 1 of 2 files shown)


src/main/scala/nodes/util/CommonSparseFeatures.scala
```diff
@@ -27,6 +27,8 @@ case class CommonSparseFeatures[T : ClassTag](numFeatures: Int) extends Estimato
     }
   }
 
+  def merge(a: Array[(T, (Int, Long))], b: Array[(T, (Int, Long))]): Array[(T, (Int, Long))] = (a ++ b).sorted(ordering.reverse).take(numFeatures)
```
etrain: pretty long line.

src/main/scala/nodes/util/CommonSparseFeatures.scala
```diff
@@ -34,7 +36,7 @@ case class CommonSparseFeatures[T : ClassTag](numFeatures: Int) extends Estimato
     val featureFrequenciesWithUniqueId = featureOccurrences.reduceByKey {
       (x, y) => (x._1 + y._1, Math.min(x._2, y._2))
     }
-    val mostCommonFeatures = featureFrequenciesWithUniqueId.top(numFeatures)(ordering).map(_._1)
+    val mostCommonFeatures = featureFrequenciesWithUniqueId.glom().map(_.sorted(ordering.reverse).take(numFeatures)).treeReduce(merge).map(_._1)
```
etrain: same here

shivaram: also a comment here on why we are doing this would be good.

etrain:
This might be a good test case for `OptimizableNode` can we come up with a rule for when its better to do this vs. the old way?

shivaram:
Isn't the difference just that we do a `treeReduce` instead of a `reduce` ? If so it should be possible to just configure `treeReduce` to behave like `reduce` when the data is small.  Pasting the `top` function implementation below. 

```
      val mapRDDs = mapPartitions { items =>
        // Priority keeps the largest elements, so let's reverse the ordering.
        val queue = new BoundedPriorityQueue[T](num)(ord.reverse)
        queue ++= util.collection.Utils.takeOrdered(items, num)(ord)
        Iterator.single(queue)
      }
      mapRDDs.reduce { (queue1, queue2) =>
          queue1 ++= queue2
          queue1
      }.toArray.sorted(ord)
```

tomerk:
Sure, I'll use that instead of the glom and sorted code.

shivaram:
Yeah and if you use the bounded queue we shouldn't see any perf degradation in terms of compute

tomerk:
Ah, spark's bounded priority queue is marked private. Should I copy the code or do something else?

shivaram:
According to http://stackoverflow.com/questions/7878026/is-there-a-priorityqueue-implementation-with-fixed-capacity-and-custom-comparato there is one in Guava. Not sure what our take on using guava is. (@etrain ?) Otherwise the answer suggested there or inserted into a sorted set is not bad

etrain:
I like the `BoundedPriorityQueue` approach. While adding a dependency on Guava for this is kind of a bummer, I'd rather that than reimplementing ourselves.

tomerk:
@etrain `BoundedPriorityQueue` is a Spark private util, not a guava class. As for the guava one mentioned in the stack overflow post: It's not a full bounded priority queue implementation, and the comments on that answer warn that it shouldn't be used for bounded priority queues:
"We are giving some consideration to deprecating MinMaxPriorityQueue, just because it tends to get misused in this way."

etrain:
Alright - what do we need? A thing that takes in an iterator over the partition and returns the top k elements by inserting/removing elements from a binary heap or something? I guess this is probably simple enough to implement ourselves but i don't want to go too deep down a rabbit hole here. Maybe there's another implementation somewhere we should be using?

tomerk:
bounded priority stuff pushed.

etrain:
Hmm - do we not have unit tests for `CommonSparseFeatures` - seems like this would be a good time to put them in to make sure that the priority queue stuff isn't breaking some edge cases. Otherwise this LGTM.

tomerk:
We already do, as part of the SparseFeaturesSuite

src/main/scala/nodes/util/CommonSparseFeatures.scala
```diff
@@ -34,8 +40,25 @@ case class CommonSparseFeatures[T : ClassTag](numFeatures: Int) extends Estimato
...
     val featureSpace = mostCommonFeatures.zipWithIndex.toMap
     new SparseFeatureVectorizer(featureSpace)
   }
+
+  /**
+   * Returns the first K elements from the input as defined by the specified implicit Ordering[T]
+   * and maintains the ordering.
+   */
+  def takeOrdered[T](input: Iterator[T], num: Int)(implicit ord: Ordering[T]): Seq[T] = {
+    val ordering = new com.google.common.collect.Ordering[T] {
```
shivaram: I don't think we explicitly depend on guava in our build.sbt and we seem to be getting it through some transitive dependencies ?

tomerk: Transitive via Spark, I'll add it as a direct dependency

tomerk:
Okay, I added an explicit guava dependency. We may have to keep an eye on this as we update Spark versions.

shivaram:
How did you pick the version number ?

tomerk:
It was what all the existing transitive dependencies on Guava had resolved to.

shivaram:
Sounds good. LGTM

PR #226, opened 2016-02-10
Convolution Speedups

etrain:
This PR speeds up Convolution by about 2.5x (on average) particularly in the case when Images are too big to fit in cache lines. The speedup comes mostly from careful loop ordering and omitting an unnecessary copy of the result.

Here are the results of running ImageBenchmarkSuite on the old code (Convolver.scala commit: 6f1f34a955b63f3347a74108513b46e245f6e823)
```
Launching sbt from sbt/sbt-launch-0.13.5.jar
[info] Loading project definition from /Users/sparks/repos/keystone/project
[info] Set current project to keystoneml (in build file:/Users/sparks/repos/keystone/)
[info] Compiling 1 Scala source to /Users/sparks/repos/keystone/target/scala-2.10/classes...
16/02/10 13:03:24 INFO ImageBenchMarkSuite: name,max(flops),median(flops),stddev(flops)
16/02/10 13:03:25 INFO ImageBenchMarkSuite: Cifar1000,15.190,9.861,3.898
16/02/10 13:03:25 INFO ImageBenchMarkSuite: Cifar10000,8.709,7.978,1.409
16/02/10 13:03:25 INFO ImageBenchMarkSuite: Cifar100,15.273,11.498,5.111
16/02/10 13:03:25 INFO ImageBenchMarkSuite: SolarFlares2,14.496,10.605,2.044
16/02/10 13:03:25 INFO ImageBenchMarkSuite: SolarFlares,7.743,7.095,1.050
16/02/10 13:03:25 INFO ImageBenchMarkSuite: ConvolvedSolarFlares,5.338,5.066,0.470
16/02/10 13:03:25 INFO ImageBenchMarkSuite: ImageNet,8.997,6.666,2.160
[info] ImageBenchMarkSuite:
[info] - Convolution Benchmarks
[info] Run completed in 2 minutes, 24 seconds.
[info] Total number of tests run: 1
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 1, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
[success] Total time: 155 s, completed Feb 10, 2016 1:03:25 PM
```

And here are the results of running on this newer code (Convolver.scala commit: cb9325accfac402d4b79bf188736b886f65f342d)
```
Launching sbt from sbt/sbt-launch-0.13.5.jar
[info] Loading project definition from /Users/sparks/repos/keystone/project
[info] Set current project to keystoneml (in build file:/Users/sparks/repos/keystone/)
[info] Compiling 1 Scala source to /Users/sparks/repos/keystone/target/scala-2.10/classes...
16/02/10 13:07:30 INFO ImageBenchMarkSuite: name,max(flops),median(flops),stddev(flops)
16/02/10 13:07:31 INFO ImageBenchMarkSuite: Cifar1000,19.814,17.254,6.138
16/02/10 13:07:31 INFO ImageBenchMarkSuite: Cifar10000,21.010,14.536,6.404
16/02/10 13:07:31 INFO ImageBenchMarkSuite: Cifar100,13.586,7.631,3.899
16/02/10 13:07:31 INFO ImageBenchMarkSuite: SolarFlares2,30.627,19.239,5.951
16/02/10 13:07:31 INFO ImageBenchMarkSuite: SolarFlares,20.362,17.903,4.420
16/02/10 13:07:31 INFO ImageBenchMarkSuite: ConvolvedSolarFlares,18.639,14.954,2.706
16/02/10 13:07:31 INFO ImageBenchMarkSuite: ImageNet,13.618,8.961,3.330
[info] ImageBenchMarkSuite:
[info] - Convolution Benchmarks
[info] Run completed in 1 minute, 3 seconds.
[info] Total number of tests run: 1
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 1, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
[success] Total time: 73 s, completed Feb 10, 2016 1:07:31 PM
```

Before merging, I'm also planning to write some extensive unit tests for Convolution to verify that we match other exact 2D convolution implementations.

The times reported by this benchmark are theoretical flops required for the convolution operation `2*(imgWidth*imgHeight)*(imgChannels*convSize^2)*(numFilters)` divided by time (in ns) = total GFLOP/s of this operation. 

The machine I'm running these on is capable of 50GFLOP/s on 1000x1000 square matrices as reported by julia using the same logic/calculation, so taking into account the irregular shape of the matrices we're multiplying in this operation, I think we're hitting pretty close to machine peak.

Further, profiling in YourKit reveals that most of the time is now being spent in the matrix multiply operations rather than `makePatches` or the old loop that copied the image out.


src/main/scala/nodes/images/Convolver.scala
```diff
@@ -101,29 +101,31 @@ object Convolver {
...
-      while (y < resHeight) {
-        chan = 0
-        while (chan < imgChannels) {
-          poy = 0
-          while (poy < convSize) {
-            pox = 0
-            while (pox < convSize) {
+
+    poy = 0
+    while (poy < convSize) {
```
etrain: Doing this is super annoying and error-prone. I'm normally allergic to macros, but this seems like the kind of thing we could make fast with meta-programming. Any ideas?

etrain:
Oh, I should note that this PR also adds support for a semi-fast Iterator interface for `Images` I think its useful to have as it will enable speedup in Pooling over our current implementation, but I'm happy to move that to a separate PR if desired.

src/main/scala/nodes/images/Convolver.scala
```diff
@@ -63,11 +63,11 @@ object Convolver {
     val convRes: DenseMatrix[Double] = imgMat * convolutions
 
     val res = new ChannelMajorArrayVectorizedImage(
-      new Array[Double](resWidth*resHeight*convolutions.cols),
+      convRes.toArray,
       ImageMetadata(resWidth, resHeight, convolutions.cols))
 
     // Now pack the convolved features into the result.
-    var x, y, chan = 0
+    /*var x, y, chan = 0
```
shivaram: can we avoid having commented code ?

src/main/scala/nodes/images/Convolver.scala
shivaram: If its something thats used multiple times we can make a utility function out of it ? But it seems like in this particular case there is some interesting logic inside ?

We could go down the macro route, but if simpler things are enough that'd be better

etrain: What I mean is - I want something that converts somehting like
```
specialFor(
  a <- 0 until amax;
  b <- 0 until bmax;
  c <- 0 until cmax;
  d <- 0 until dmax;
  e <- 0 until emax) {

//do something with a,b,c,d,e
}
```

Into a loop like this at compile time.

shivaram: If you are okay with a function call overhead and handling return types somehow you can do that with a simple util function which looks like. 
```
specialFor(0 until amax, 0 until bmax, fn: (a: T, b: T) => U) = {
   while (avar < amax) {
     bvar = 0
     while(bvar < bmax) {
       fn(avar, bvar)
       avar = avar + 1 
    }
     avar = avar + 1
  }
}
```

src/main/scala/utils/images/Image.scala
```diff
@@ -183,6 +183,15 @@ case class ChannelMajorArrayVectorizedImage(
     channelIdx + x * metadata.numChannels + y * metadata.xDim * metadata.numChannels
   }
 
+  override def vectorToImageCoords(v: Int): Coordinate = {
```
shivaram: Would be great to have some unit tests for each of these

src/main/scala/utils/images/Image.scala
```diff
@@ -251,8 +274,36 @@ trait VectorizedImage extends Image {
   override def put(x: Int, y: Int, channelIdx: Int, newVal: Double) = {
     putInVector(imageToVectorCoords(x, y, channelIdx), newVal)
   }
+
+  def vectorToImageCoords(v: Int): Coordinate = ???
+
+  val coord: Coordinate = new Coordinate(0,0,0)
+
+  def iter(): Iterator[CoordinateValue] = new Iterator[CoordinateValue] {
```
shivaram: Would be good to document that the reference to CoordinateValue shouldn't be modified or saved by the caller.

src/main/scala/nodes/images/Convolver.scala
etrain: the function call might get inlined in our case, but who knows.

There's also this: https://github.com/nativelibs4java/Scalaxy/tree/master/Loops

Which is the macro-rewriting optimized for loop thing that someone else has done for us. I might try both - but after getting decent convolver unit tests put together.

src/main/scala/utils/images/Image.scala
shivaram: Also calling iter from multiple threads will cause bad things to happen from what i can see as `coord` is updated by all the calls to `vectorToImageCoords` -- is that right ?

src/test/scala/nodes/util/MatrixMultiplySuite.scala
```diff
@@ -0,0 +1,64 @@
+package utils
+
+import breeze.linalg.{DenseVector, SparseVector, DenseMatrix}
+import org.scalatest.FunSuite
+import pipelines.Logging
+
+class MatrixMultiplySuite extends FunSuite with Logging {
+  test("Vector Matrix mult bench") {
```
shivaram: Do you want to keep this file or was this just for getting an idea of system limits ?

etrain:
Currently this is broken. I am comparing output visually and by pixel to an equivalent method in scipy - will let you know when things are better.

etrain:
Also - FWIW the bug doesn't come at the expense of speedup - it's can likely be fixed with a slightly different ordering in `makePatches`

etrain:
After some work on this, I
1. Added a new interface for the Convolver to make it easier to use - end users shouldn't care about the underlying ordering of, e.g. the patches. 
2. Added a unit test that does an *exact* comparison with scipy - this was really useful in debugging as it caught some loop ordering mistakes, etc. Also, we *really* need to figure out a sane multi-cube story. At this point I'd *really* be into picking a single dimension ordering and sticking to it, regardless of what it costs us in performance. The main things I ran into were actually related to this, our (weird as hell) BGR ordering of channels, and the fact that some implementations of convolution *flip* the convolution kernels across all dimensions before applying them (aka scipy).
3. Changed the output image format of the convolver. This shouldn't affect anything since everything takes an `Image` but it will probably make the SymmetricRectifier slower in the common case. We can save this update for a future PR.

After all that, the new performance numbers are looking pretty great (now ~3x improvement over existing), though, and tests pass:

```
[info] Compiling 1 Scala source to /Users/sparks/repos/keystone/target/scala-2.10/classes...
16/02/16 10:22:57 INFO ImageBenchMarkSuite: name,max(flops),median(flops),stddev(flops)
16/02/16 10:22:57 INFO ImageBenchMarkSuite: Cifar1000,29.626,28.161,5.789
16/02/16 10:22:57 INFO ImageBenchMarkSuite: Cifar10000,31.319,26.334,9.207
16/02/16 10:22:57 INFO ImageBenchMarkSuite: Cifar100,17.042,15.454,5.199
16/02/16 10:22:57 INFO ImageBenchMarkSuite: SolarFlares2,34.523,29.457,2.879
16/02/16 10:22:57 INFO ImageBenchMarkSuite: SolarFlares,28.383,25.231,6.830
16/02/16 10:22:57 INFO ImageBenchMarkSuite: ConvolvedSolarFlares,22.928,20.947,4.638
16/02/16 10:22:57 INFO ImageBenchMarkSuite: ImageNet,18.248,14.629,6.122
```

etrain:
Pushed a few fixes. I still need to add some unit tests for the image iterator, which I'll do shortly. But @shivaram @tomerk the bulk of the code should be reviewable.

etrain:
Alright - I think this is ready to be looked at.

build.sbt
```diff
@@ -21,7 +21,7 @@ libraryDependencies ++= Seq(
   "edu.arizona.sista" % "processors" % "3.0" classifier "models",
   "org.slf4j" % "slf4j-api" % "1.7.2",
   "org.slf4j" % "slf4j-log4j12" % "1.7.2",
-  "org.scalatest" %% "scalatest" % "1.9.1" % "test",
+  "org.scalatest" %% "scalatest" % "2.2.6" % "test",
```
shivaram: is this safe ? why did we need this ?

src/main/scala/nodes/images/Convolver.scala
```diff
@@ -46,6 +46,107 @@ class Convolver(
...
+      imgInfo.xDim,
+      imgInfo.yDim,
+      imgInfo.numChannels,
+      whitener,
+      normalizePatches,
+      varConstant)
+  }
+
+  /**
+    * Flip the image such that flipImage(im)(x,y,z) = im(im.metadata.xDim-x,im.metadata.yDim-y,im.metadata.numChannels-z)
```
shivaram: This should go in ImageUtils or something like that ?

build.sbt
etrain: This was so that I could run a single test from the command line, which was very useful for benchmarking, e.g.:
```
sbt/sbt "test-only nodes.images.ImageBenchMarkSuite -- -z Convolution" > `git rev-parse HEAD`.log
```

etrain: Further, I don't think it does anything bad to the tests. I will double-check though.

src/main/scala/nodes/images/Convolver.scala
etrain: Sure.

src/main/scala/utils/images/Image.scala
```diff
@@ -152,10 +152,12 @@ case class ImageMetadata(xDim: Int, yDim: Int, numChannels: Int)
 case class ByteArrayVectorizedImage(
     vectorizedImage: Array[Byte],
     override val metadata: ImageMetadata) extends VectorizedImage {
-  override def imageToVectorCoords(x: Int, y: Int, channelIdx: Int): Int = {
+  def imageToVectorCoords(x: Int, y: Int, channelIdx: Int): Int = {
     channelIdx + y*metadata.numChannels + x*metadata.yDim*metadata.numChannels
   }
 
+  def vectorToImageCoords(v: Int): Coordinate = ???
```
tomerk: I think this is supposed to be deleted

tomerk: What does the fact that this isn't implemented make ByteArrayVectorizedImage unusable for?

src/main/scala/utils/images/ImageUtils.scala
```diff
@@ -46,6 +46,20 @@ object ImageUtils extends Logging {
   }
 
   /**
+    * Writes image to file `fname`
+    * @param fname Destination filename.
+    * @param in
+    * @param scale
```
tomerk: Can we add the comments from `imageToBufferedImage` to this?

src/test/python/images/pyconv.py
```diff
@@ -0,0 +1,25 @@
...
+import csv
+import numpy as np
+from scipy.misc import imread, imsave
+from scipy.signal import convolve
+
+#This script convolves an image in python and its output is used in
+#the Convolver unit tests to ensure that convolver output matches
+#an equivalent python call.
+
+def main():
```
tomerk: Does this get run w/ `sbt/sbt test`?

build.sbt
etrain: Hmm.. it does seem to break the tests - anything that runs with a SparkContext is affected by this - we end up with a bunch of errors like this:
```

src/test/python/images/pyconv.py
etrain: no - it already generated the file in `src/test/resources/images/convolve.gantrycrane.csv`

Any suggestions about better ways to organize it?

src/main/scala/utils/images/Image.scala
etrain: I can implement it - it would make it unusable for anything using the `.iter()` method, which is nothing right now.

build.sbt
shivaram: My guess is there is some class version confusion ? This is happening in the executor while its trying to deserialize a task. Also the missing class is a Scala collection -- can you check if the scalatest pulls in some other scala version ?

src/test/python/images/pyconv.py
tomerk: Hmm that makes sense. Could you just put a comment saying that you already executed it (and the command you used)?

build.sbt
etrain: Alright - fighting with class path hell doesn't seem like a productive use of time right now. I'm going to revert and worst case we can factor out the convolution benchmarks into a separate class.

src/main/scala/utils/images/Image.scala
etrain: Done.

src/main/scala/utils/images/ImageUtils.scala
etrain: Done.

src/test/python/images/pyconv.py
etrain: Done.

src/main/scala/utils/images/Image.scala
etrain: Done.

etrain:
Alright - PR comments addressed - feel free to take another pass @shivaram @tomerk

tomerk:
LGTM

etrain:
If it looks good feel free to merge.

PR #227, opened 2016-02-11
Node-level optimization generalization

tomerk:
Allows node-level optimization rules to construct pipelines instead of just returning a single node.

The code in it's current form is very convoluted but the unit tests pass. Suggestions for how to improve it would be appreciated.

### DIFF
--- a/src/main/scala/workflow/DefaultOptimizer.scala
+++ b/src/main/scala/workflow/DefaultOptimizer.scala
@@ -7,7 +7,7 @@ import workflow.AutoCacheRule.GreedyCache
  */
 object DefaultOptimizer extends Optimizer {
   protected val batches: Seq[Batch] =
-    Batch("DAG Optimization", FixedPoint(Int.MaxValue), EquivalentNodeMergeRule) ::
+    Batch("Common Sub-expression Elimination", FixedPoint(Int.MaxValue), EquivalentNodeMergeRule) ::
     Batch("Node Level Optimization", Once, new NodeOptimizationRule) ::
     Batch("Common Sub-expression Elimination", FixedPoint(Int.MaxValue), EquivalentNodeMergeRule) ::
       Nil
@@ -18,8 +18,9 @@ object DefaultOptimizer extends Optimizer {
  */
 class AutoCachingOptimizer(strategy: AutoCacheRule.CachingStrategy = GreedyCache()) extends Optimizer {
   protected val batches: Seq[Batch] =
-    Batch("DAG Optimization", FixedPoint(100), EquivalentNodeMergeRule) ::
+    Batch("Common Sub-expression Elimination", FixedPoint(Int.MaxValue), EquivalentNodeMergeRule) ::
     Batch("Node Level Optimization", Once, new NodeOptimizationRule) ::
+    Batch("Common Sub-expression Elimination", FixedPoint(Int.MaxValue), EquivalentNodeMergeRule) ::
     Batch("Auto Cache", Once, new AutoCacheRule(strategy)) ::
       Nil
 }
--- a/src/main/scala/workflow/DefaultOptimizer.scala
+++ b/src/main/scala/workflow/DefaultOptimizer.scala
@@ -7,8 +7,9 @@ import workflow.AutoCacheRule.GreedyCache
  */
 object DefaultOptimizer extends Optimizer {
   protected val batches: Seq[Batch] =
-    Batch("DAG Optimization", FixedPoint(100), EquivalentNodeMergeRule) ::
+    Batch("DAG Optimization", FixedPoint(Int.MaxValue), EquivalentNodeMergeRule) ::
     Batch("Node Level Optimization", Once, new NodeOptimizationRule) ::
+    Batch("Common Sub-expression Elimination", FixedPoint(Int.MaxValue), EquivalentNodeMergeRule) ::
       Nil
 }
 
--- a/src/main/scala/workflow/NodeOptimizationRule.scala
+++ b/src/main/scala/workflow/NodeOptimizationRule.scala
@@ -8,13 +8,15 @@ import org.apache.spark.rdd.RDD
  * @param samplePerPartition The number of items per partition to look at when optimizing nodes
  */
 class NodeOptimizationRule(samplePerPartition: Int = 3) extends Rule {
-  override def apply[A, B](plan: Pipeline[A, B]): Pipeline[A, B] = {
-    val instructions = WorkflowUtils.pipelineToInstructions(plan)
+  case class OptimizationState(
+    optimizedInstructions: Seq[Instruction],
+    unoptimizedInstructionIndexToOptimizedIndex: Int => Int,
+    registers: Map[Int, InstructionOutput],
+    numPerPartitionPerNode: Map[Int, Map[Int, Int]]
+  )
 
-    // First, figure out which instructions we should actually optimize and execute.
-    // Execute the set of all instructions that are parents of an optimizable node transform or fit instruction.
-    // And, keep those that do not depend on the runtime input of a fit pipeline
-    val instructionsToOptimize = instructions.zipWithIndex.collect {
+  private def getInstructionsToOptimize(instructions: Seq[Instruction]): Set[Int] = {
+    instructions.zipWithIndex.collect {
       case (TransformerApplyNode(tIndex, _), index)
         if instructions(tIndex).isInstanceOf[OptimizableTransformer[_, _]] => index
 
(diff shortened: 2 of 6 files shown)


src/main/scala/workflow/DefaultOptimizer.scala
```diff
@@ -5,8 +5,9 @@ package workflow
  */
 object DefaultOptimizer extends Optimizer {
   protected val batches: Seq[Batch] =
-    Batch("DAG Optimization", FixedPoint(100), EquivalentNodeMergeRule) ::
+    Batch("DAG Optimization", FixedPoint(Int.MaxValue), EquivalentNodeMergeRule) ::
```
etrain: Is this dangerous? Could we end-up in a (near) infinite loop if we have a really bad looking DAG?

tomerk: Given it's a DAG, assuming the equivalent node merging isn't altering the plan when there's nothing to merge it should be O(max depth of DAG). I do need to check the node merging to make sure though.

tomerk: Just checked this, it should be fine.

etrain:
@tomerk let's make sure we get the fixes from the node-level optimization experiments in here.

tomerk:
@etrain I've separated some of the things into methods and removed much of the mutability

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -5,96 +5,341 @@ import org.apache.spark.rdd.RDD
 /**
  * Node-level optimization, such as selecting a Linear Solver
  *
- * @param sampleFraction The fraction of the RDD to use for operations
+ * @param samplePerPartition The number of items per partition to look at when optimizing nodes
  */
-class NodeOptimizationRule(sampleFraction: Double = 0.01, seed: Long = 0) extends Rule {
+class NodeOptimizationRule(samplePerPartition: Int = 3) extends Rule {
```
etrain: Can we call it `samplesPerPartition`

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -5,96 +5,341 @@ import org.apache.spark.rdd.RDD
...
+  /**
+   * Optimizes the transformer being applied by the instruction at the given index.
+   * Outputs a new optimization state.
+   *
+   * @param os The previous optimization state
+   * @param transformerApplyNode
+   * @param index
+   * @return
+   */
+  private def optimizeTransformer(os: OptimizationState, transformerApplyNode: TransformerApplyNode, index: Int)
```
etrain: Yikes - os is a bad name for this. how about `optState` or something? `os.copy` made me think that we were copying files and I thought something was very wrong..

tomerk: Yeah I realized it was a suspicious name after the fact....

src/main/scala/workflow/NodeOptimizationRule.scala
tomerk: sure thing!

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -5,96 +5,341 @@ import org.apache.spark.rdd.RDD
...
+
+  /**
+   * Get the set of instructions that we need to check for optimizations at.
+   *
+   * This is any instruction that either applies an optimizable transformer,
+   * or fits an optimizable estimator,
+   * and is not supposed to be executed on test data.
+   */
+  private def getInstructionsToOptimize(instructions: Seq[Instruction]): Set[Int] = {
+    instructions.zipWithIndex.collect {
```
etrain: Should we explicitly handle the base case here? `.collect` is one I never really use because its meaning is overloaded in Spark - but it's appropriate here!

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -5,96 +5,341 @@ import org.apache.spark.rdd.RDD
...
+   * @param index
+   * @return
+   */
+  private def executeInstruction(optimizationState: OptimizationState, instruction: Instruction, index: Int)
+  : OptimizationState = {
+    // Get the dependencies from the registers and run the instruction on them (caching RDD outputs)
+    val registers = optimizationState.registers
+    val deps = instruction.getDependencies.map(registers)
+    val instructionOutput = instruction.execute(deps) match {
+      case RDDOutput(rdd) => RDDOutput(rdd.cache())
```
etrain: Looks like we never clean up after ourselves here. I realize it's a data sample, but is it possible to add a hook to the end of the optimization procedure such that things we cached here are unpersisted cleanly? Totally understandable if it's not easy.

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -5,96 +5,341 @@ import org.apache.spark.rdd.RDD
...
+      case dep if numPerPartitionPerNode.contains(dep) => numPerPartitionPerNode(dep)
+    }
+    val newNumPerPartitionPerNode = if (numPerPartition.nonEmpty) {
+      numPerPartitionPerNode + (index -> numPerPartition.get)
+    } else {
+      numPerPartitionPerNode
+    }
+
+    // Create the new state
+    optimizationState.copy(registers = newRegisters, numPerPartitionPerNode = newNumPerPartitionPerNode)
```
etrain: "Immutability changes everything." - Pat Helland :+1:

src/main/scala/workflow/NodeOptimizationRule.scala
tomerk: Yeah no problem, I already have such a hook in the autocacherule and I forgot about putting one here. will do!

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -5,96 +5,341 @@ import org.apache.spark.rdd.RDD
...
+      case EstimatorOutput(oe: OptimizableEstimator[a, b]) => {
+        val spliceEndpoint = Pipeline.SOURCE - 1
+
+        // Produce the optimized estimator fit instructions that we then
+        // need to splice into the pipeline
+        val dataSample = inputs.head.asInstanceOf[RDD[a]]
+        val pipeToSplice = oe.optimize(dataSample, numPerPartition).apply(dataSample)
+        val initialInstructionsToSplice = WorkflowUtils.pipelineToInstructions(pipeToSplice)
+        val dataSampleIndices = initialInstructionsToSplice.indices.filter {
+          initialInstructionsToSplice(_) == SourceNode(dataSample)
```
etrain: nested `_` inside `()` always make me worry about scope - can we be explicit here with `.filter(inx => `

src/main/scala/workflow/WorkflowUtils.scala
```diff
@@ -159,4 +159,99 @@ object WorkflowUtils {
...
+          }
+        )
+    }
+
+    val oldToNewIndexMapping = (i: Int) => i - offsets.getOrElse(i, 0)
+
+    (newInstructions, oldToNewIndexMapping)
+  }
+
+  def disconnectAndRemoveInstructions(dependencyReplacement: Map[Int, Int], instructions: Seq[Instruction])
```
etrain: Can we add some scaladoc explaining what the arguments are and what this is doing? This is a pretty heavily used procedure.

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -5,96 +5,341 @@ import org.apache.spark.rdd.RDD
...
+    val inputIndices = transformerApplyNode.inputs
+
+    val numPerPartition = os.numPerPartitionPerNode(inputIndices.head)
+    val inputs = inputIndices.map(os.registers).collect {
+      case RDDOutput(rdd) => rdd
+    }
+
+    // Optimize the transformer if possible
+    os.registers(tIndex) match {
+      case TransformerOutput(ot: OptimizableTransformer[a, b]) => {
```
etrain: Is there a reason you're using lower-case type names here? Makes it a little surprising/hard to read.

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -5,96 +5,341 @@ import org.apache.spark.rdd.RDD
...
+   * @param transformerApplyNode
+   * @param index
+   * @return
+   */
+  private def optimizeTransformer(os: OptimizationState, transformerApplyNode: TransformerApplyNode, index: Int)
+  : OptimizationState = {
+    val tIndex = transformerApplyNode.transformer
+    val inputIndices = transformerApplyNode.inputs
+
+    val numPerPartition = os.numPerPartitionPerNode(inputIndices.head)
```
etrain: Is there an assumption about input ordering here? As in, the first RDD input to the transformer is the training one, or is it just a case of "we're optimizing the transformer for the types of input its going to be fed, so let's pick the first one?"

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -5,96 +5,341 @@ import org.apache.spark.rdd.RDD
...
+
+        os.copy(
+          optimizedInstructions = spliceResult._1,
+          unoptimizedInstructionIndexToOptimizedIndex = intermediateIndexMapping andThen spliceResult._2
+        )
+      }
+      case _: TransformerOutput => os
+      case _ => throw new ClassCastException("TransformerApplyNode dep wasn't pointing at a transformer")
+    }
+  }
```
etrain: Modulo a few style nits, this whole function is *much* more readable.

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -5,96 +5,341 @@ import org.apache.spark.rdd.RDD
...
+
+    // Optimize the estimator if possible
+    os.registers(estIndex) match {
+      case EstimatorOutput(oe: OptimizableEstimator[a, b]) => {
+        val spliceEndpoint = Pipeline.SOURCE - 1
+
+        // Produce the optimized estimator fit instructions that we then
+        // need to splice into the pipeline
+        val dataSample = inputs.head.asInstanceOf[RDD[a]]
+        val pipeToSplice = oe.optimize(dataSample, numPerPartition).apply(dataSample)
```
etrain: Here, and above (in the OtimizableTransformer), you may want to space out this line of code by itself and add a comment that says "Actually optimize the {transformer,estimator} based on a datasample and statistics about the data partitions."

src/main/scala/workflow/NodeOptimizationRule.scala
tomerk: Yes, you have to use lower-case type names for generic type pattern matching to work.

src/main/scala/workflow/NodeOptimizationRule.scala
tomerk: will do!

src/main/scala/workflow/WorkflowUtils.scala
tomerk: will do!

src/main/scala/workflow/NodeOptimizationRule.scala
etrain: Also - I'm not sure I follow the return types here. The result of `oe.optimize()` is an `Estimator`. What is the apply method of an `Estimator`?

src/main/scala/workflow/NodeOptimizationRule.scala
tomerk: The assumption is that if a transformer takes multiple RDDs, we expect them to be fully zippable / copartitioned, and the output to not mess w/ the partitioning. Do you think that's a safe assumption? Either way I'll add a comment.

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -5,96 +5,341 @@ import org.apache.spark.rdd.RDD
...
+        }
+
+        os.copy(
+          optimizedInstructions = partialOptimizedInstructions,
+          unoptimizedInstructionIndexToOptimizedIndex = partialOldIndexToNewIndex
+        )
+      }
+
+      case EstimatorOutput(oe: OptimizableLabelEstimator[a, b, l]) => {
+        val dataSpliceEndpoint = Pipeline.SOURCE - 1
```
etrain: Oh man, these three lines are pretty worrisome. Why are these the right numbers? Where is this order dependence coming from? Can you at least put a comment here?

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -5,96 +5,341 @@ import org.apache.spark.rdd.RDD
...
+        val labelSpliceEndpoint = Pipeline.SOURCE - 2
+        val spliceEndpoint = Pipeline.SOURCE - 3
+
+        // Get the instructions to splice into the optimized pipeline
+        val dataSample = inputs(0).asInstanceOf[RDD[a]]
+        val labelSample = inputs(1).asInstanceOf[RDD[l]]
+        val pipeToSplice = oe.optimize(dataSample, labelSample, numPerPartition)
+          .apply(dataSample, labelSample)
+        val initialInstructionsToSplice = WorkflowUtils.pipelineToInstructions(pipeToSplice)
+        val dataSampleIndices = initialInstructionsToSplice.indices.filter {
```
etrain: same comment here as above. will not make again.

src/main/scala/workflow/NodeOptimizationRule.scala
tomerk: The return type of `oe.optimize()` isn't `Estimator`, it's `RDD[A] => Pipeline[A, B]`

tomerk: So what is happening here is that we're constructing a pipeline but w/ the sampled RDD instead of the true RDD, then we're removing the uses of the sampled RDD to produce instructions for a partial pipeline, that is the optimization of the `estimator`

src/main/scala/workflow/NodeOptimizationRule.scala
tomerk: kk, done

src/main/scala/workflow/NodeOptimizationRule.scala
tomerk: No, we actually specifically don't want to have a base case so that anything that doesn't match ends up not in the set. Collect is kind of like a mix of map and filter in that sense. If this is too scala magicky let me know and I'll replace it.

src/main/scala/workflow/NodeOptimizationRule.scala
tomerk: Will do! Numbers and order don't matter, the only things that matters are that they are different and don't refer to any existing instruction.

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -204,28 +213,33 @@ class NodeOptimizationRule(samplePerPartition: Int = 3) extends Rule {
...
 
-        os.copy(
+        optState.copy(
           optimizedInstructions = partialOptimizedInstructions,
           unoptimizedInstructionIndexToOptimizedIndex = partialOldIndexToNewIndex
         )
       }
 
       case EstimatorOutput(oe: OptimizableLabelEstimator[a, b, l]) => {
+        // We start by setting constants to refer to endpoint 'ids'.
```
etrain: Good :)

src/main/scala/workflow/NodeOptimizationRule.scala
```diff
@@ -340,6 +354,12 @@ class NodeOptimizationRule(samplePerPartition: Int = 3) extends Rule {
       }
     }
 
+    // Unpersist any RDDs we cached
+    finalOptimization.registers.values.foreach {
```
etrain: Terrific!

src/main/scala/workflow/WorkflowUtils.scala
```diff
@@ -192,6 +192,19 @@ object WorkflowUtils {
     (newInstructions, oldToNewIndexMapping)
   }
 
+  /**
+   * This method first replaces all dependencies on specific instructions with a different int, effectively
```
etrain: Nice, thanks! This stuff is super useful from a maintainability perspective.

etrain:
Alright, this LGTM. Thanks a bunch @tomerk !

Issue #213, opened 2016-01-25
Optimize Weighted Block Solver for Low # of classes


etrain:
The current parallelization strategy is over classes, so in (e.g. binary classification) the weighted block solver is slow at scale.

cc @ericmjonas

shivaram:
Fixed by #228

PR #228, opened 2016-02-13
WeightedLeastSquares solver for small number of classes

shivaram:
The `BlockWeightedLeastSquares` solver in Keystone has a restriction where the number of partitions must be equal to the number of classes. While this works for large class problems like ImageNet it reduces parallelism for smaller problems.

This PR introduces a `PerClassWeightedLeastSquares` implementation that solves the weighted least squares objective [1] for a single class in parallel and does this sequentially `k` times for a `k` class problem. The advantage here is that the parallelism is not limited by number of classes and hence this will be faster when `k` is small. 

[1] https://onlinecourses.science.psu.edu/stat501/node/352

cc @ericmjonas @etrain


shivaram:
Fixes #213

shivaram:
So I tested this out using `RandomPatchCifar` on the AMP Spark cluster. The good news is that the accuracy that we get from this vs. the old solver remains the same (at 4k and 40k features with mixWt 0 and 0.25).

However the BlockWeightedLeastSquares was faster at both 4k and 40k features. I guess this is partly to be expected given the algorithms. The trade-off is as follows (for say 4k features, with block size 4k)
1. In the old BWLS we shuffle data to fit in 10 partitions, compute A.tA and the solve for W using 10 partitions. So `time_taken = time_to_shuffle + n*d^2/10+ d^3`
2. In the new code, for each class we compute A.tA in parallel using 192 partitions and then solve for `W` locally. So `time taken = 10 * n*d^2/192 + 10 * d^3`. 
Of course as you might notice I am only including compute time here and not communication. For a small dataset like CIFAR using 192 partitions for `At * A` adds more communication. 

tldr; the code is correct and will probably help for the binary classification that @ericmjonas is doing.  But building the optimizer on top of this needs some more work.

shivaram:
I refactored the code to create a separate general purpose solver for the traditional definition of weighted least squares i.e. a system like `W = (X.t * (diag(B) * X) + \lambda * I) \ X.t * (diag(B) .* Y)`
This is used in the per-class weighted solver but can also be used to implement IRWLS.

The one catch is that I didn't want to expose this as an estimator as it doesn't really make sense. So I put it in `nodes.learning.internal`. Let me know if you have any thoughts about this.

etrain:
This LGTM - we can attack optimizability in a future PR, I think.

etrain:
Thanks @shivaram !

Issue #230, opened 2016-02-17
Broadcast W, b in RandomCosineFeatures


shivaram:
This affects performance when we have large number of random features and was found by @ericmjonas and @Vaishaal 

cc @etrain

etrain:
As @Vaishaal also notes, we could simply generate these on the Executor side and save the network bandwidth. Relies on seeding the RNG exactly the same on all nodes and duplicates work, but should shave some end-to-end runtime.

etrain:
Ignoring the prposal above (generate on the executor side) 

@shivaram - can you describe what the issue with not explicitly broadcasting is? Are we sending the whole matrix once per executor or something? I understand that W and b are getting serialized with the task, but are the tasks not getting broadcast now?

shivaram:
So from what I remember - we needed to do this in the suncast case and @tomerk added it https://github.com/ericmjonas/suncast/commit/90df2b8b4cb98198bad052a3c2c38081ca50f2f7 (which is partly why I assigned it to him)

I think the root cause was something either around task deserialization taking too long or the the task broadcast being repeated when do somethings iteratively. @tomerk / @ericmjonas do you have any ideas on why you did this ?

Irrespective I don't think its a bad idea to use a broadcast variable when its really large amount of data. It lets you control the lifecycle much better.

tomerk:
Totally forgot I wrote that. At the time I was probably under the impression that the matrix was getting transferred over the network once per task, but that might not have been the case.

ericmjonas:
Doesn't this behavior also potentially change between spark 1.3 and 1.6?
Suncast only targets 1.6 now

shivaram:
Ok I don't mind closing this issue as won't fix - but I'll feel better about it if we do two things
(a) Run suncast with Spark 1.6 after switching it back to the non-broadcasted version
(b) Make sure the TIMIT pipeline with 200k features using the `Pipeline.gather` + random cosine features doesn't suffer in performance

tomerk:
Hmm either way we should probably add the node @Vaishaal wrote for generating the random features matrix on the executors. I just need to know what changes still need to be made to turn that node into a PR if you guys want me to.

shivaram:
@Vaishaal 's code is at https://raw.githubusercontent.com/Vaishaal/ckm/whitening/keystone_pipeline/src/main/scala/nodes/stats/SeededCosineRandomFeatures.scala

tomerk:
[...]

PR #231, opened 2016-02-19
Added seed cosine random features

tomerk:
issue #230 
@Vaishaal

### DIFF
--- a/src/main/scala/nodes/stats/CosineRandomFeatures.scala
+++ b/src/main/scala/nodes/stats/CosineRandomFeatures.scala
@@ -17,16 +17,18 @@ import workflow.Transformer
  * Kernel trick to allow Linear Solver to learn cosine interaction terms of the input
  */
 class CosineRandomFeatures(
-    val W: DenseMatrix[Double], // should be numOutputFeatures by numInputFeatures
-    val b: DenseVector[Double]) // should be numOutputFeatures by 1
+    @transient val W: DenseMatrix[Double], // should be numOutputFeatures by numInputFeatures
+    @transient val b: DenseVector[Double]) // should be numOutputFeatures by 1
     extends Transformer[DenseVector[Double], DenseVector[Double]] {
 
   require(b.length == W.rows, "# of rows in W and size of b should match")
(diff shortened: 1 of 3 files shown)


shivaram:
Can't we just modify the existing RandomCosineFeatures ?

tomerk:
I think it makes sense to have them separate, since it's a tradeoff. Regenerating the whole matrix for each partition (or each single-item apply) instead of serializing and broadcasting it comes with its own different cost.

shivaram:
I'm playing devils advocate here partly, but is there really a trade-off ? i.e. do we have any numbers which show how say as the matrix size, num machines increases the cost of sending data vs. regenerating it changes ?

Because the problem with adding more nodes for each trade-off is that somebody (human or optimizer) needs to choose among them. And I worry without an idea of what is expensive / cheap this won't every be better than trial / error.

tomerk:
That's a fair point, and I don't have any numbers like that for the RDD apply case. (Although trivially if you're looking to do very low-latency single-item applies generating it only once at the start makes much more sense than regenerating it for each item). 

@Vaishaal and @ericmjonas you two seem to be the ones stress-testing cosine random features currently, do you two have any sense of this?

Vaishaal:
So yes I agree with @tomerk that regenerating for every time will be prohibitively expensive. We only generate the kernel matrix from the seed every partition. And number of partitions should be << number of items so it seems like a good tradeoff..

etrain:
I'm a little confused. Do we still even want the seeded one? @shivaram @tomerk

shivaram:
No lets remove the seeded one for now. We'll run regression tests and then add one if we can figure out a regime where its useful

tomerk:
Maybe I should just make a separate PR for the broadcasted cosine random features then? (because this is specifically the seeded random cosines PR) @shivaram @etrain

etrain:
Sure.

tomerk:
See #245

etrain:
Closing this in favor of #245 - if we decide we really want this functionality we can bring it back.

PR #232, opened 2016-02-19
Replace LLH computation with a while loop

### LINKED ISSUE #174: Make Scala GMM faster

shivaram:
This PR is an attempt to make the Scala GMM faster. I tried a bunch of changes, but the that was most useful was replacing the log-likelihood computation with a while loop. Some numbers from before vs. after 
```
Dataset : 1M points with 80 dimensions. OMP_NUM_THREADS is set to 1
Time is in seconds
Num centers, Old time, With this PR time
64, 383.481, 266.444
32, 214.411, 149.455
16, 90.724, 66.057
```

So we get around 25-30% improvement in performance from this change.

Fixes #174

### DIFF
--- a/src/main/scala/nodes/learning/GaussianMixtureModelEstimator.scala
+++ b/src/main/scala/nodes/learning/GaussianMixtureModelEstimator.scala
@@ -110,10 +110,8 @@ case class GaussianMixtureModelEstimator(
     while ((iter < maxIterations) && costImproving && largeEnoughClusters) {
       /* E-STEP */
 
-      /*
-      compute the squared malhanobis distance for each gaussian.
-      sq_mal_dist(i,j) || x_i - mu_j||_Lambda^2.
-      */
+      // compute the squared malhanobis distance for each gaussian.
+      // sq_mal_dist(i,j) || x_i - mu_j||_Lambda^2.
       val sqMahlist = (XSq * gmmVars.map(0.5 / _).t) - (X * (gmmMeans :/ gmmVars).t) +
           (DenseMatrix.ones[Double](numSamples, 1) * (sum(gmmMeans :* gmmMeans :/ gmmVars, Axis._1).t :* 0.5))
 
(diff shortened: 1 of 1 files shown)


src/main/scala/nodes/learning/GaussianMixtureModelEstimator.scala
```diff
@@ -110,10 +110,8 @@ case class GaussianMixtureModelEstimator(
...
       /* E-STEP */
 
-      /*
-      compute the squared malhanobis distance for each gaussian.
-      sq_mal_dist(i,j) || x_i - mu_j||_Lambda^2.
-      */
+      // compute the squared malhanobis distance for each gaussian.
+      // sq_mal_dist(i,j) || x_i - mu_j||_Lambda^2.
       val sqMahlist = (XSq * gmmVars.map(0.5 / _).t) - (X * (gmmMeans :/ gmmVars).t) +
           (DenseMatrix.ones[Double](numSamples, 1) * (sum(gmmMeans :* gmmMeans :/ gmmVars, Axis._1).t :* 0.5))
```
etrain: This seems like an unnecessary allocation. Since we are just adding some (mean^2/variance)/2 vector to each row, can we use breeze's broadcasting for this? e.g.
`val sqMahlist = ((XSq * (0.5 :/ gmmVars).t - (X * (gmmMeans :/ gmmVars).t)).apply(*, ::) + (sum(gmmMeans :* gmmMeans :/ gmmVars, Axis._1) :+ 0.5)
`
This ends up taking the same amount of time in my tests, but is close.

etrain:
One note, otherwise looks good to me.

src/main/scala/nodes/learning/GaussianMixtureModelEstimator.scala
shivaram: Done - I split it into two lines and did a in-place sum using breeze's broadcasting.

etrain:
For sanity's sake can we get the timing with this PR in its current state?

src/main/scala/nodes/learning/GaussianMixtureModelEstimator.scala
```diff
@@ -110,37 +110,41 @@ case class GaussianMixtureModelEstimator(
...
-      /*
-      compute the squared malhanobis distance for each gaussian.
-      sq_mal_dist(i,j) || x_i - mu_j||_Lambda^2.
-      */
-      val sqMahlist = (XSq * gmmVars.map(0.5 / _).t) - (X * (gmmMeans :/ gmmVars).t) +
-          (DenseMatrix.ones[Double](numSamples, 1) * (sum(gmmMeans :* gmmMeans :/ gmmVars, Axis._1).t :* 0.5))
+      // compute the squared malhanobis distance for each gaussian.
+      // sq_mal_dist(i,j) || x_i - mu_j||_Lambda^2.
+      val sqMahlist = (XSq * (0.5 :/ gmmVars).t) - (X * (gmmMeans :/ gmmVars).t)
+      sqMahlist(*, ::) += (sum(gmmMeans :* gmmMeans :/ gmmVars, Axis._1) :+ 0.5)
```
etrain: oh, this sshould be `:* 0.5`, not `:+ 0.5`

shivaram:
Timing numbers with this change
```
66.2448
149.145
270.772
```

etrain:
Thanks - LGTM!

PR #233, opened 2016-02-22
Fixes to get builds to work again

tomerk:
@shivaram @etrain

### DIFF
--- a/src/main/scala/utils/images/Image.scala
+++ b/src/main/scala/utils/images/Image.scala
@@ -282,7 +282,7 @@ trait VectorizedImage extends Image {
 
   def vectorToImageCoords(v: Int): Coordinate
 
-  val coord: Coordinate = new Coordinate(0,0,0)
+  @transient lazy val coord: Coordinate = new Coordinate(0,0,0)
 
   /**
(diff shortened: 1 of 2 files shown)


etrain:
Did you run the full suite of tests locally on this? In particular - something which serializes `Image` would be good?

src/main/scala/utils/images/Image.scala
```diff
@@ -282,7 +282,7 @@ trait VectorizedImage extends Image {
 
   def vectorToImageCoords(v: Int): Coordinate
```
shivaram: Can we make this method and the `coord` object `private` ? Right now `coord` is only used by this method, so it gives us some assurance it won't be accessed outside of it.

etrain: Yeah, I think that's a fine idea. We haven't done this much yet, but given
that this thing is potentially "dangerous" I guess we should go ahead and
do this.

shivaram:
Did the tests pass locally ?

shivaram:
Alright - since you confirmed this offline, I'm merging this now.

PR #234, opened 2016-03-05
Kernel solver

Vaishaal:
@shivaram let me know if this looks correct. I'm storing the full model in memory and broadcasting, do you know of a more effective way to do this.


shivaram:
cc @stephentu @rolloff

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -0,0 +1,273 @@
...
+import breeze.numerics._
+import org.apache.spark.rdd.RDD
+import org.apache.spark.SparkConf
+import org.apache.spark.SparkContext
+import org.apache.spark.SparkContext._
+import utils._
+
+
+// Trait to be implemented by Kernel functions
+trait KernelGenerator[T] {
```
shivaram: If we think of kernels as a feature transformation, then `KernelGenerator` can be a Transformer ? @sparks Its fine for Transformers to run jobs right ?

src/main/scala/nodes/learning/KernelModel.scala
```diff
@@ -0,0 +1,78 @@
...
+ **/
+
+class KernelBlockModel(
+  models: Broadcast[Array[DenseMatrix[Double]]],
+  blockSize: Int,
+  lambdas: Array[Double],
+  trainData: RDD[DenseVector[Double]],
+  kernelGen: KernelGenerator[DenseVector[Double]],
+  nTrain: Option[Int] = None
+)
```
shivaram: The keystone-ml project follows for more part the Style guide laid out at https://github.com/databricks/scala-style-guide -- specifically see the `Spacing and indentation` section there

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -0,0 +1,273 @@
...
+    testMat: RDD[T],
+    trainMat: RDD[T],
+    blockIdxs: Seq[Int]): RDD[DenseVector[Double]]
+
+}
+
+object SparseLinearKernel extends KernelGenerator[(Array[Long], Array[Double])]
+  with Serializable {
+
+  private def sparseDotProduct(
```
shivaram: I'm not sure this is how we deal with Sparse data in KeystoneML (check the Amazon pipeline for example). I think we can leave out this Sparse linear kernel in this PR ?

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -0,0 +1,273 @@
...
+      }.toArray)
+    }
+
+    val Kblockblock = blockDataArray.par.map { case (lhsInds, lhsValues) =>
+      new DenseVector(blockDataArray.map { case (rhsInds, rhsValues) =>
+        sparseDotProduct(lhsInds, lhsValues, rhsInds, rhsValues)
+      }.toArray)
+    }.toArray
+
+    // val Kblockblock = blockDataArray.map { case (lhsInds, lhsValues) =>
```
shivaram: General rule of thumb: get rid of all commented code.

src/main/scala/nodes/learning/KernelRidgeRegression.scala
```diff
@@ -0,0 +1,315 @@
...
+
+        // materialize the new model
+        newModel.count()
+
+        models.unpersist(true)
+        models = newModel
+
+        val modelUpdateTime = timeElasped(newModelBegin)
+
+        println(s"EPOCH_${pass}_BLOCK ${block} took " + timeElasped(blockBegin) + " seconds")
```
shivaram: Another guideline is to not use `println` but `logInfo`

src/main/scala/utils/KernelUtils.scala
```diff
@@ -0,0 +1,440 @@
+package utils
+
+import nodes.stats._
+import org.apache.spark.rdd.RDD
+import org.apache.spark.broadcast.Broadcast
+import breeze.linalg._
+import scala.collection.mutable.HashMap
+import scala.reflect.ClassTag
+
+object KernelUtils extends Serializable {
```
shivaram: This file shouldn't be there. Almost all of these util functions should be in MatrixUtils in Keystone (else you can add them there)

src/main/scala/nodes/learning/KernelRidgeRegression.scala
```diff
@@ -0,0 +1,315 @@
...
+    //
+    // Cache this as n*k should be small
+    // TODO: Should we zero mean the labels ?
+    val labelsMat = labels.mapPartitions { part =>
+      Iterator.single(KernelUtils.rowsToMatrix(part))
+    }.cache()
+    labelsMat.count
+    val labelsRPM = RowPartitionedMatrix.fromMatrix(labelsMat)
+
+    val kWs = lambdas.map { l =>
```
shivaram: Looks like `kWs` is not used anymore ?

etrain:
One thing i would like to try and get right here is interfaces. E.g. The sparseLinearKernel is of type `(Array[Long], Array[Double])` when probably it should be of type SparseVector? I am not sure we are (ever) going to be running 2e9 by 2e9 kernel solves, so maybe this is OK? Also, is the right return value for, e.g. `generateKernelTestBlock` actually `RDD[DenseVector[Double]` or is it a `RowPartitionedMatrix`? In the only use of that function in this pull request, the first thing we do is convert each partition to a DenseMatrix, which could probably be handled locally more efficiently.

Shivaram brings up a point about whether the KernelGenerator should be a `Transformer`. It would be good to figure out how to fit this into the `Estimator`/`Transformer` framework. The fact that there's a separate method for Train/Test indicates to me that maybe we should have one method that takes Train and is an Estimator that when fit returns a `Transformer` that can be applied to test. I haven't thought through the workflow of these things as closely as you guys, but hopefully that provides some guidance. Taking a pass over the rest of the code now.

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -0,0 +1,273 @@
...
+import org.apache.spark.SparkConf
+import org.apache.spark.SparkContext
+import org.apache.spark.SparkContext._
+import utils._
+
+
+// Trait to be implemented by Kernel functions
+trait KernelGenerator[T] {
+
+  // TODO(stephentu): these should both be just one function
```
etrain: see my note about Estimator vs. Transformer.

src/main/scala/nodes/learning/KernelGenerator.scala
etrain: Yeah, +1 for leaving it out if we don't have this particular interface figured out. Would be really nice to get in general, though!

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -0,0 +1,273 @@
...
+    val trainDotProd = trainMat.map {x =>
+      (x.t*x).toDouble
+    }
+    val testDotProd = testMat.map {x =>
+      (x.t*x).toDouble
+    }
+
+    // b x d block of training data
+    val blockDataArray = trainMat.zipWithIndex.filter{ case (vec, idx) =>
+      blockIdxs.contains(idx)
```
etrain: this could get slow if someone chooses a big block size. You might want to make blockIdxs a "Set" if order is unimportant - or convert locally to a set for this kidn of check.

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -0,0 +1,273 @@
...
+class GaussianKernelGenerator(gamma: Double)
+  extends KernelGenerator[DenseVector[Double]]
+  with Serializable {
+
+  def generateKernelTestBlock(
+    testMat: RDD[DenseVector[Double]],
+    trainMat: RDD[DenseVector[Double]],
+    blockIdxs: Seq[Int]): RDD[DenseVector[Double]]= {
+
+    println("testMat.count=" + testMat.count +
```
etrain: If you need this it should be a `logInfo` but I don't think we need them - especially since they trigger jobs.

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -0,0 +1,273 @@
...
+    // Dot product of rows of X with each other
+    val trainDotProd = trainMat.map {x =>
+      (x.t*x).toDouble
+    }
+    val testDotProd = testMat.map {x =>
+      (x.t*x).toDouble
+    }
+
+    // b x d block of training data
+    val blockDataArray = trainMat.zipWithIndex.filter{ case (vec, idx) =>
```
etrain: Not sure if `.zipWithIndex` is what we want here - if `trainMat` were converted to a `RowPartitionedMatrix` here there would be enough metadata to do this. Maybe @shivaram has some thoughts.

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -0,0 +1,273 @@
...
+      blockIdxs.contains(idx)
+    }.map(x=> x._1).collect()
+    val blockData = KernelUtils.rowsToMatrix(blockDataArray)
+    assert(blockData.rows == blockIdxs.length)
+    val blockDataBC = trainMat.context.broadcast(blockData)
+
+    // <xi,xj> for i in [nTest], j in blockIdxs
+    val blockXXT = testMat.mapPartitions { itr  =>
```
etrain: Thinking in terms of abstraction here - the point here is that we're generating a subset of the (all-pairs) Kernel matrix. Should the interface for the `KernelGenerator` take a kernel function which is of type `(Vector, Vector) => Double`? And then we could subclass or something to handle this case where we have an efficient `(Matrix, Matrix) => Matrix` way of doing it?

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -0,0 +1,273 @@
...
+    trainBlockDotProdBC.unpersist()
+    kTestBlock
+  }
+
+  // Generate a block of size n x b of the kernel matrix
+  def generateKernelTrainBlock(
+    mat: RDD[DenseVector[Double]],
+    blockIdxs: Seq[Int]): (RDD[DenseVector[Double]], DenseMatrix[Double]) = {
+
+    assert(KernelUtils.isSorted(blockIdxs)) // limitation for now
```
etrain: Can you elaborate on where this limitation comes from?

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -0,0 +1,273 @@
...
+
+    println("mat.count=" + mat.count + ", blockSize=" + blockIdxs.length)
```
etrain: Again, extra job/ `println`

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -0,0 +1,273 @@
...
+    // <xi,xj> for i in [nTrain], j in blockIdxs
+    val kBlock = mat.zipPartitions(matDotProd) { case (iterMat, iterDotProds) =>
+      val bd = blockDataBC.value
+      val bmdp = blockMatDotProdBC.value
+
+      val xMat = KernelUtils.rowsToMatrix(iterMat.toArray)
+      val vecDotProds = DenseVector(iterDotProds.toArray)
+      val XXTTerm = xMat * bd.t
+
+      //// non-vectorized version
```
etrain: dead code

src/main/scala/nodes/learning/KernelModel.scala
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * @param models different kernel models for different lambdas
+ * @param blockSize blockSize to split data before applying transformations
+ * @param lambdas  lambdas corresponding to the models
+ * @param trainData the training Data
+ * @param kernelGen the kernel generator
+ **/
+
+class KernelBlockModel(
+  models: Broadcast[Array[DenseMatrix[Double]]],
```
etrain: Would rather not have a `Broadcast` here - if we get some new RDD with some new `SparkContext` we should hopefully be able to apply the model there (though maybe not if we really need `TrainData` below). 

I guess it would be good to have some utilities for loading one of these given a SparkContext and a pointer to the `TrainData` on HDFS.

src/main/scala/nodes/learning/KernelModel.scala
```diff
@@ -0,0 +1,78 @@
...
+  blockSize: Int,
+  lambdas: Array[Double],
```
etrain: This multi-lambda in a Transformer thing is a little unconventional, because it means to get a single prediction out someone needs to chain on another transformer. It's probably fine but we should decide how we want these interfaces to look.

src/main/scala/nodes/learning/KernelModel.scala
```diff
@@ -0,0 +1,78 @@
...
+  trainData: RDD[DenseVector[Double]],
+  kernelGen: KernelGenerator[DenseVector[Double]],
+  nTrain: Option[Int] = None
```
etrain: My guess is that to train one of these things you must have computed nTrain already, no? If so, probably best to make this *not* optional.

src/main/scala/nodes/learning/KernelModel.scala
shivaram: Yeah we can also get rid of the multi-lambda stuff for now -- The hyper parameter tuning API changes should help us incorporate this ?

src/main/scala/nodes/learning/KernelModel.scala
```diff
@@ -0,0 +1,78 @@
...
+)
+extends Transformer[DenseVector[Double], Array[DenseVector[Double]]]
+{
+  override def apply(in: RDD[DenseVector[Double]]): RDD[Array[DenseVector[Double]]] =  {
+    val trainSize: Int = nTrain.getOrElse(trainData.count().toInt)
+    val testSize = in.count()
+    val numBlocks = math.ceil(trainSize.toDouble/blockSize).toInt
+    // Only evaluate first model for now
```
etrain: I guess you probably only want to take in a single lambda then?

etrain: And therefore I guess you only take in a model.

src/main/scala/nodes/learning/KernelModel.scala
```diff
@@ -0,0 +1,78 @@
...
+    // TODO: MULTI LAMBDA
+    val oneModel = models.value(0)
+    val model = trainData.context.broadcast(oneModel)
```
etrain: Probably cleaner if you used the `.context` from `in`?

src/test/scala/nodes/learning/KernelModelSuite.scala
```diff
@@ -0,0 +1,37 @@
...
+import edu.berkeley.cs.amplab.mlmatrix.RowPartitionedMatrix
+import nodes.stats.StandardScaler
+import org.apache.spark.SparkContext
+import org.scalatest.FunSuite
+import pipelines.{LocalSparkContext, Logging}
+import utils.{MatrixUtils, Stats, KernelUtils}
+
+class KernelModelSuite extends FunSuite with LocalSparkContext with Logging {
+
+  test("KernelModel XOR test") {
```
etrain: I like it!

etrain:
Alright - I took a pass with a bunch of style/structure comments. Before you do a ton of coding it might make sense to just sketch out the interface changes you're thinking about and we can talk about them here?

Also, from a maintainability perspective, it would be really good to get a reference to a book chapter or paper that describes exactly the problem you're solving in the scaladoc and then be consistent with those variable names, etc. For example, I looked at 7.4 of Murphy's book to page this into my brain this AM. For the main algorithm - I guess the BCD kernel paper is on arXiv now? http://arxiv.org/pdf/1602.05310v1.pdf

src/main/scala/nodes/learning/KernelModel.scala
```diff
@@ -0,0 +1,78 @@
...
+    val numClasses = oneModel.cols
+    val testMatrix = KernelUtils.rowsToMatrix(in)
+
+
+    /* Initially all predictions are 0 */
+    var predictions = testMatrix.map(x => DenseMatrix.zeros[Double](x.rows, x.cols))
```
Vaishaal: Whoops this is a bug! it should be x.rows, numClasses

src/main/scala/nodes/learning/KernelGenerator.scala
Vaishaal: It's unclear to me how a KernelGenerator is a Transformer, since it behaves fundamentally difference for train and test data

shivaram: Isn't it just a transformer that is (a) initialized with the trainData and then it can be applied to train and test data ? i.e. if K is our kernel we run `K(x, data)` where data is the training data and `x` can be either train or test ?

Vaishaal: i guess i don't know what the one element equivalent would be. Since a transformer in my head is valid for a single data point, but a kernel isn't really well defined for it?

Vaishaal: i guess the first element could be forced to be the entire data hm

Vaishaal: i think my mental model of a kernel isn't really a transformation, since the whole point is that we can't explicitly store the transformation. And its not really analogous to any of the other transformers we have (SIFT/CKN/RandomCosine), idk it seems just a bit contrived. But I can do it.

shivaram: The first element is that element dot trainingData -- i.e. With a linear kernel lets say you have a full dataset Y (m x d) and a train data set X (n x d). Then the full kernel is Y * X^T (size is m x n)

Now if I only get 1 row of Y, say y1 then the kernel is y1 * X^T -- which is of size 1 x n

Does this make sense ?

Vaishaal: yeah

shivaram: I think we need to store the transformation for it to be the same across train, test ? I agree as generally defined, a Kernel is K(x, y) = <value> i.e. it is defined on two data points. What we are doing here is fixing one of them, so we are defining K1(x) = K(x, trainRDD). Now you can replace trainRDD with random gaussian and you get RF. Thats my mental model fwiw

Vaishaal: so the correct way to generate the kernel matrix for training data is to pass in trainMat into the constructor and then trainMat into the apply function?

shivaram: yeah - that'll work fine. if you want to do something more efficient for the case were its K(train, train) then you can check the RDD `id` and if they are the same RDD do something more clever etc.

Vaishaal: also what do we do about the second return value?

Vaishaal: do we just bite the bullet and recompute that?

shivaram: we have two choices (a) create a data structure called `KernelBlock` that stores two things a RDD and a DenseMatrix (which is K_block_block) (b) we throw it away and re-create it inside the algorithm given the KernelBlock. It shouldn't be very expensive and this was a performance optimization that doesn't hold up very well to abstraction

Vaishaal: ok I'll do b.

src/main/scala/nodes/learning/KernelGenerator.scala
Vaishaal: For now I think we should leave the KernelGenerator as is, since we don't have many other kernels yet, and maybe we can come up with a pattern once we have multiple kernels...

src/main/scala/nodes/learning/KernelGenerator.scala
Vaishaal: @shivaram

shivaram: I'm not sure if this limitation still exists. Need to look through the code in a bit more detail. Basically if I remember right the limitation came from needing to intersect the block indexes with the matrix row indexes  -- @stephentu do you remember where this comes from ?

stephentu: I think this simplifies having one more lookup table-- there is really no reason to prefer any particular ordering of columns within a block, so you might as well take the sorted order. Another way to say this is that `blockIdxs` should be `Set[Int]` instead of `Seq[Int]`.

Vaishaal: Oh ok lets just have a set, since I convert it to a set anyway.

Vaishaal: Hmm it seems you can only index RowPartitionedMatrices by Range objects,
which have to be sorted...

src/main/scala/utils/MatrixUtils.scala
```diff
@@ -69,6 +69,35 @@ object MatrixUtils extends Serializable {
   }
 
   /**
+   *  Converts a RDD of DenseVector to a RDD of DenseMatrix such that each partition
+   *  of  DenseVectors are stacked into a rows to make  DenseMatrix
+   */
+
+  def rowsToMatrix[T: ClassTag] (in: RDD[DenseVector[T]]) : RDD[DenseMatrix[T]]= {
+    in.mapPartitions { part =>
+      Iterator.single(rowsToMatrix(part))
```
shivaram: Use `rowsToMatrixIter` here

src/main/scala/utils/MatrixUtils.scala
```diff
@@ -69,6 +69,35 @@ object MatrixUtils extends Serializable {
...
+    }
+  }
+  /**
+   * Converts a  RDD of DenseMatrix where each matrix is in one partition into 
+   * an RDD of DenseVectors separating the matrix row wise.  Undefined behavior
+   * if used on RDD of  DenseMatrix not generated  from above function
+   */
+
+  def matrixToRows [T: ClassTag] (in: RDD[DenseMatrix[T]]) : RDD[DenseVector[T]]= {
+    in.mapPartitions { part =>
```
shivaram: you can just call `matrixToRowArray` inside the `mapPartitions` ?

Vaishaal: I used it in a couple places, and I wanted to follow DRY :)

src/main/scala/nodes/learning/KernelBlockModel.scala
```diff
@@ -0,0 +1,74 @@
...
+ * @param models different kernel models for different lambdas
+ * @param blockSize blockSize to split data before applying transformations
+ * @param lambdas  lambdas corresponding to the models
+ * @param in the training Data
+ * @param kernelGen the kernel generator
+ * @param nTrain number j training examples
+ **/
+
+class KernelBlockModel(
+  model: DenseMatrix[Double],
```
shivaram: So the model won't fit in a single matrix if the data is large due to Java array size limits. For example if you have 2M examples and 1000 classes, the size of the model is `n*k` which is 2 billion elements and very close to the java array limit of 2^31 - 1.  We should pass in a blocked model (similar to BlockLinearMapper)

src/main/scala/nodes/learning/KernelBlockModel.scala
```diff
@@ -0,0 +1,74 @@
...
+  blockSize: Int,
+  lambdas: Array[Double],
+  kernelGen: KernelGenerator,
```
shivaram: So having made the kernelGenerator a transformer we shouldn't need to take this in as a part of the model ? Shouldn't it already be a part of the pipeline ?

src/main/scala/nodes/learning/KernelRidgeRegression.scala
```diff
@@ -0,0 +1,343 @@
...
+import utils._
+
+
+import edu.berkeley.cs.amplab.mlmatrix.RowPartitionedMatrix
+import edu.berkeley.cs.amplab.mlmatrix.util.{Utils => MLMatrixUtils}
+
+import scala.util.Random
+
+class KernelRidgeRegression(
+  kernelBlockGenerator: KernelGenerator,
```
shivaram: Similar comment as in the kernelModel - I think we can make the `kernel` transformation a part of the pipeline and then this solver is just a BCD based solver for `Kx = b` where `K` is the input to this estimator.

@etrain Does this sound good to you ?

etrain: Yeah, I think I like this. My mental model:
`KernelEstimator: Estimator[DenseVector[Double], Matrix]`  (basically does nothing but pass along a reference to the input RDD), .fit returns a KernelGenerator
`KernelGenerator: Transformer[DenseVector[Double], Matrix]` (does the work of forming the Kernel matrix - note that this can be applied to training or test data)
Then the output of the KernelGenerator is passed to BlockLeastSquaresEstimator.
So the whole pipeline is as simple as 
`Preprocessing andThen (KernelEstimator, train) andThen (BlockLeastSquaresEstimator, train, labels)`

We can wrap this all in a `Pipeline` object (or a function that takes an RDD and returns a `Pipeline` or something) so that the whole thing can be chained in the same way we chain a vanilla linear solver to the end of a pipeline.

Does this make sense to both of you @shivaram @Vaishaal

shivaram: Partly - but not completely. There are a couple of things which aren't different IMHO from your description

- In theory don't need to do any estimation for the KernelGenerator. The Generator is just a transformer which takes in a RDD[DenseVector] (i.e. training data) in its constructor.  I don't know if that is elegant to do in Keystone
- The LinearSystemSolver is different from LeastSquares -- The former solves `Ax = b` while the latter solves `|| Ax - b||_2`.  We definitely want to use the former after applying the kernel.

- Then your pipeline can be something like 
```
val trainData = preProcessing(inputData) 
val kernel = new RBFKernelGenerator(trainData)
val predictorPipeline = preProcessing andThen kernel andThen new LinearSystemSolver(blockSize, lambda)
```
- So the only weird thing about the code above is that the RBFGenerator can't be chained using `andThen`. I guess if you make it an `Estimator` then it might be possible ?

etrain: @Vaishaal and I talked about this on the whiteboard briefly. I think I basically understand the math difference now. We can clarify on a whiteboard at some point. We should be a little careful with our use of the word "solves" since i think that's overloaded. Basically what we're saying is if A is a linear system with identical equations and unknowns (as in the case where A is the Kernel matrix, we want to find x which is an exact solution if one exists). We still use BCD to find this solution, rather than directly inverting A, but we are not minimizing the least squares loss.
In the case where A is not square (either A is overdetermined or underdetermined), we're going to find the X which minimizes the least-squares loss as you state. 

On the KernelGenerator bit, I think from a syntax perspective it's better to have this be an estimator. It also makes it easier to track usages of this RDD for auto-caching, etc. Also, a kernel function can be viewed as a degenerate kNN which is something I'd think of as an Estimator :). This gets the pipeline back to (roughly) how i wrote it before.

Vaishaal: Morally/mathematically its the same thing since if you minimize ||Ax - b|| for a PD A, you will recover A^{-1}b. Just implementation wise, since we know the kernel is PD we want to do the linear system solve. Am I right @shivaram?

shivaram: Its equivalent in the sense that you will get the same answer but its worse in terms of conditioning / computation costs. To make this a bit more clear in the least squares case we are solving `x = (A^T * A)^-1 (A^T * b)`
So here the condition number of `A` is squared and we need to spend time to compute `A^T * A`

But if A is a square matrix (like a kernel) then we can solve a linear system `x = A^-1 * b`. Now you can use the least squares method for the square matrix as well but that'll unnecessarily make the condition number worse and we'll need to spend a lot of FLOPs computing `A^T * A` (n^3 for a n*n matrix). 

TLDR is that for square matrices you can solve the linear system directly instead of using the least squares method

Vaishaal: okay after thinking about this a bit I think the kernelBlockGenerator should still be passed into krr.scala (maybe not kernelModel). There are certain optimizations I do/computations I cache by calling KernelBlockGenerator inside the loop. Also it makes the cacheKernel argument kind of confusing since the user could have materialized the kernel outside the call into KRR. Maybe we can talk about this in person?

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -22,9 +22,8 @@ abstract class KernelGenerator extends Transformer[DenseVector[Double], DenseVec
   /* Cached KernelBlockBlock (used for training), and corresponding
    * blockIdx
    */
-  var cachedBlockBlock:Option[(DenseMatrix[Double], Seq[Int])] = None
 
-  def apply(testMat: RDD[DenseVector[Double]], blockIdxs: Seq[Int], same: Boolean = false, cache: Boolean = false): RDD[DenseVector[Double]]
+  def apply(testMat: RDD[DenseVector[Double]], blockIdxs: Seq[Int], same: Boolean = false): RDD[DenseVector[Double]]
```
etrain: We ultimately want this returning a KernelMatrix, right?

src/main/scala/nodes/learning/KernelMatrix.scala
```diff
@@ -0,0 +1,78 @@
...
+  /* Return a n x idxs.size sub matrix (every entry in RDD is a row)
+   * @param idxs the column indexes of your kernel matrix you want
+   * */
+  def apply(idxs: Seq[Int]):RDD[DenseMatrix[Double]]
+
+  def apply(idxs1: Seq[Int], idxs2: Seq[Int]): RDD[DenseMatrix[Double]]
+
+}
+
+class BlockedKernelMatrix (kernelGen: KernelGenerator, data: RDD[DenseVector[Double]],  train: Boolean, cacheKernel: Boolean) extends KernelMatrix
```
etrain: long line.

src/main/scala/nodes/learning/KernelMatrix.scala
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+trait KernelMatrix {
+
+  /* Return a n x idxs.size sub matrix (every entry in RDD is a row)
+   * @param idxs the column indexes of your kernel matrix you want
+   * */
+  def apply(idxs: Seq[Int]):RDD[DenseMatrix[Double]]
+
+  def apply(idxs1: Seq[Int], idxs2: Seq[Int]): RDD[DenseMatrix[Double]]
```
etrain: I assume this is subsampling rows as well as columns? Would be nice to change cariable names to be like `rowIdxs` and `colIdxs`, to make it more explicit.

etrain: and change the previous column only case to be `colIdxs` explicitly.

src/main/scala/nodes/learning/KernelGenerator.scala
Vaishaal: I thought you create a KernelMatrix by passing in a KernelGenerator and data

src/main/scala/nodes/learning/KernelMatrix.scala
```diff
@@ -0,0 +1,78 @@
...
+
+}
+
+class BlockedKernelMatrix (kernelGen: KernelGenerator, data: RDD[DenseVector[Double]],  train: Boolean, cacheKernel: Boolean) extends KernelMatrix
+{
+  val cache = HashMap.empty[(Seq[Int], Seq[Int]),RDD[DenseMatrix[Double]]]
+
+  def apply(idxs: Seq[Int]):RDD[DenseMatrix[Double]] = {
+    val kBlock =
+    if (cache contains ((idxs, Seq.empty[Int]))) {
```
etrain: spacing here is kind of weird, maybe indent this block? or put the `if` on the same line as `=`

etrain: Minor nit - the indexing for matrices is `(row,column)` so it feels strange to me that this Map key is `(columns,())` I would expect it to be `((),columns)`

Vaishaal: technically this is a symmetric matrix always so it shouldn't matter right?

Vaishaal: wait nvm

src/main/scala/nodes/learning/KernelMatrix.scala
```diff
@@ -25,49 +25,52 @@ trait KernelMatrix {
 
 }
 
-class BlockedKernelMatrix (kernelGen: KernelGenerator, data: RDD[DenseVector[Double]],  train: Boolean, cacheKernel: Boolean) extends KernelMatrix
+class BlockedKernelMatrix (kernelGen: KernelGenerator,
```
shivaram: This is still not the right style. Please see https://github.com/databricks/scala-style-guide#spacing-and-indentation (especially the one about `For classes whose header doesn't fit in a single line`)

src/main/scala/nodes/learning/KernelMatrix.scala
```diff
@@ -0,0 +1,78 @@
...
+      cache((idxs, Seq.empty[Int]))
+    } else {
+    val kBlock = toMatrix(kernelGen(data, idxs, train))
+      if (cacheKernel) {
+        kBlock.cache()
+        cache += ((idxs, Seq.empty[Int])  -> kBlock)
+      }
+    kBlock
+    }
+    kBlock
```
etrain: Looks like you can have this whole function just be the if expression?

src/main/scala/nodes/learning/KernelMatrix.scala
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  def apply(idxs1: Seq[Int], idxs2: Seq[Int]):RDD[DenseMatrix[Double]] = {
+    /* Check if the block we are looking for is in cache */
+    val kBlockMatrix =
+    if (cache contains (idxs1, idxs2)) {
+      cache((idxs1, idxs2))
+    } else {
+      val kBlock =
+      if (cache contains (idxs1, Seq.empty[Int])) {
```
etrain: Same thing on formatting here.

src/main/scala/nodes/learning/KernelMatrix.scala
```diff
@@ -25,49 +25,52 @@ trait KernelMatrix {
 
 }
 
-class BlockedKernelMatrix (kernelGen: KernelGenerator, data: RDD[DenseVector[Double]],  train: Boolean, cacheKernel: Boolean) extends KernelMatrix
+class BlockedKernelMatrix (kernelGen: KernelGenerator,
+                           data: RDD[DenseVector[Double]],
+                           train: Boolean,
```
shivaram: Can we detect this `same` automatically by comparing RDD ids ? Or will that not work ?

src/main/scala/nodes/learning/KernelMatrix.scala
```diff
@@ -0,0 +1,78 @@
...
+    } else {
+      val kBlock =
+      if (cache contains (idxs1, Seq.empty[Int])) {
+        cache((idxs1, Seq.empty[Int])).flatMap(MatrixUtils.matrixToRowArray(_))
+      } else {
+        kernelGen(data, idxs1, false)
+      }
+
+      val kBlockBlock =
+      toMatrix(kBlock.zipWithIndex.filter{ case (vec, idx) =>
```
etrain: @shivaram may have opinions here, but what are our assumptions on the row layout here? Are they that the index created by `.zipWithIndex` is going to give us an appropriate matrix row ID?

src/main/scala/nodes/learning/KernelGenerator.scala
etrain: Basically we need an `Estimator` that takes in training data and produces a `Transformer` from `RDD[DenseVector[Double]]` to `KernelMatrix`, or `SimilarityMatrix`, right?

etrain: Or am I not looking for that in the right spot?

etrain:
Ok - for right now I'm just focused on the `KernelMatrix` API/implementation because I think the other changes we talked about yesterday are still to come. Modulo my comment about `zipWithIndex` and some style nits I am basically cool with this.

src/main/scala/nodes/learning/KernelMatrix.scala
Vaishaal: Yeah we can probably do that.

src/main/scala/nodes/learning/KernelMatrix.scala
Vaishaal: Yeah I'll defer to @shivaram but this is how I imagined we could do it. (This is how KernelGenerator generates a block anyway).

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * Return a column block of the kernel test matrix
+   *
+   * The kernel test matrix is a |test| x |train| matrix K_testtrain.
+   * This function returns the columns in blockIdxs of K_testtrain.
+   *
+   * WARNING: Observe the order that test comes first
+   */
+
+  def apply(testMat: RDD[DenseVector[Double]], blockIdxs: Seq[Int], same: Boolean = false): RDD[DenseVector[Double]] = {
```
etrain: Actually, something definitely feels weird to me here. I'm a little surprised to see "same" mattering at this level, as in, it almost shouldn't matter if two matrices are the same or different, computing a block of the kernel matrix is simply computing the kernel function on all the "test" data and the relevant ids in the "train" data - b.t.w. I'm not sure "train" and "test" are the right things to be using here - that's a common case for this but you can imagine this being validation or whatever - this is a more general operator than train vs. test.

Vaishaal: I guess this was just an optimization for the symmetric case. When you want to generate the kernel matrix for your training data. I can remove it and recompute it every time.

Vaishaal: If see same is only used to avoid recomputing trainDotProd

etrain: Ah, ok - i see that now - yeah, in that case I think what @shivaram suggested with the RDD ids might be a little cleaner. If you've got this info sitting around anyway, probably a good idea not to recompute it.

Vaishaal: yeah didn't know about rdd ids. Fixed now

src/main/scala/nodes/learning/KernelGenerator.scala
Vaishaal: Do we want KernelMatrix to be an estimator? I don't see it fitting in either as a Transformer or Estimator. Since its sort of like a datastructure

shivaram: Just keeping it as a data structure is fine.

Vaishaal:
If this looks good I'll start plumbing the linear system stuff and then we should be done?

src/main/scala/nodes/learning/KernelGenerator.scala
shivaram: But I think the KernelGenerator should be a transformer that goes from data to KernelMatrix ? Then the LinearSystemSolver can be an estimator that takes in KernelMatrix and generates a BlockLinearModel.

src/main/scala/nodes/learning/KernelMatrix.scala
shivaram: Yeah this is fine. It runs an extra job which seems wasteful. Son one thing I've done in the past is to have an index RDD cached once so you don't need to do this per block (assuming the indexes don't change).

src/main/scala/nodes/learning/KernelGenerator.scala
Vaishaal: Currently as implemented KernelMatrix uses KernelGenerator to return parts of the KernelMatrix, you want me to flip that?

Vaishaal: https://github.com/amplab/keystone/blob/kernelSolver/src/main/scala/nodes/learning/KernelRidgeRegression.scala#L74 

Currently I generate the KernelMatrix at the top of KernelRidgeRegression and then I'm going to wrap the rest of the class LinearSystemSolve

shivaram: yeah so the KernelMatrix can still take a function (it can even the same generator object) in the constructor.  But the nice thing about what I said before is that it will give us some modularity where we can say

`val kernelRidgePipeline = new GaussianKernelGenerator() andThen new LinearSystemSolver()` and `kernelRidgePipeline` can be `fit` and then applied to test data etc. i.e. matching the usual keystone workflow. The only difference in this case is that the intermediate datastructure is a `KernelMatrix` instead of `RDD[DenseVector[Double]]` and thus we also use the type system effectively. i.e. if you did `andThen new BlockWeightedLeastSquares` it will fail.

Vaishaal: wait if you make it a Transformer won't it expect RDD[KernelMatrix] as an output type, whereas KernelMatrix already nests an RDD[DenseVector] inside it?

shivaram: Hmm - this sucks a little bit. I don't like the transformer API is so specific that it expects things to work on `RDD[T] => RDD[U]` and `T=>U` and you can't have a custom data structure. I need to think a bit more what we can do given this restriction.

@etrain any thoughts ?

Vaishaal: Ok yeah this sucks. How about for this cut. We have KernelBlockModel as a transformer from DenseVector[RDD] to DenseVector[RDD]  and have it take in the KernelGenerator as before. Trying to fix this properly has caused  much headache :(

etrain: Let's be careful about adding more `FunctionNode`s - if anything I'd like to figure out how to get rid of them (or find a way to make them first class citizens). 

I think the problem is that `Transformers` and `Estimators` are typed at the item level, and we are conceptualizing `KernelGenerator` at the collection level.

One simple fix, potentially, is to have `KernelMatrix extends RDD[DenseVector[Double]]` and have it also have the block generator interface we talked about. 

Then, if we have:
`val kernelRidgePipeline = Identity andThen (new GaussianKernelGenerator(), trainData) andThen (new LinearSystemSolver(), trainData, trainLabels)`

The types are as follows:

```scala
GaussianKernelGenerator: Estimator[DenseVector[Double], DenseVector[Double]] {
  def fit(in: RDD[DenseVector[Double]]): Kernelizer = ???
}

class Kernelizer: Transformer[DenseVector[Double],DenseVector[Double]] {
  override def apply(in: RDD[DenseVector[Double]]): KernelMatrix = ??? //Yields the whole kernel matrix
  def apply(in: DenseVector[Double]): DenseVector[Double] = ??? //Do this for a single point.
}

class KernelMatrix extends RDD[DenseVector[Double]] {
  def getIndices(x: Seq[Int]): Block = ??? //Block generating logic
}

class LinearSystemSolver extends LabelEstimator[DenseVector[Double],DenseVector[Double],DenseVector[Double]] {
  def fit(in: RDD[DenseVector[Double]): LinearMapper = {
    in match {
     case x: KernelMatrix => //do something with x.getIndices here and generate a LinearMapper
     case _ => //Eh, maybe we don't implement this for now?
   }
  }
}
```

Does this look sane or am I missing something @Vaishaal @shivaram ?

shivaram: I think it looks sane -- I haven't thought through it completely. But it doesn't achieve the static-time type safety I was hoping for where by the LinearSystem system solver doesn't need to that `in match` stuff.

Vaishaal:
*ping*

shivaram:
Sorry for the delay - I'll get to this after May 20

Vaishaal:
sounds good!

etrain:
Alright - closing this since we got this functionality in #284.

Issue #235, opened 2016-03-09
Fix tests for PCA, LBFGS, and Image benchmarks.


etrain:
We have a few issues with tests right now:

1. PCA is non-deterministic.
2. LBFGS is non-deterministic.
3. Benchmarks are hogging time, ram, and producing a ton of console output.

I have ideas about how to fix all three.

etrain:
Closed by #236

PR #236, opened 2016-03-10
Fixing build non-determinism. Addresses #235

etrain:
I fixed the three issues described in #235 and have run the previously non-deterministic tests a large number of times with no issues.

Quick summary:
1. ImageBenchMarkSuite: I removed a big benchmark that was hogging a lot of time and lowered the log level for some fine grained statistics. This particular test was also forcing the heap space for our build to be too high.
2. PCA - Made the approximate PCA algorithm seedable and seeded the unit tests. 
3. LBFGS - Seeded random number generator.

In addition, there were a couple of other tests that use randomness that I went ahead and made seeded.

cc @shivaram @tomerk

### DIFF
--- a/src/main/scala/nodes/learning/ApproximatePCA.scala
+++ b/src/main/scala/nodes/learning/ApproximatePCA.scala
@@ -3,8 +3,10 @@ package nodes.learning
 import breeze.linalg._
 import breeze.numerics._
 import breeze.stats._
+import breeze.stats.distributions.{Gaussian, ThreadLocalRandomGenerator, RandBasis}
 import com.github.fommil.netlib.LAPACK._
 import edu.berkeley.cs.amplab.mlmatrix.util.QRUtils
+import org.apache.commons.math3.random.MersenneTwister
 import org.apache.spark.rdd.RDD
 import org.netlib.util.intW
 import pipelines.Logging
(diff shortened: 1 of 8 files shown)


etrain:
Addresses #235

src/main/scala/nodes/learning/ApproximatePCA.scala
```diff
@@ -62,10 +64,11 @@ object ApproximatePCAEstimator {
    * @param q Number of iterations to use in the computation.
    * @return An n x (k+p) matrix that approximates A
    */
-  def approximateQ(A: DenseMatrix[Double], l: Int, q: Int): DenseMatrix[Double] = {
+  def approximateQ(A: DenseMatrix[Double], l: Int, q: Int, seed: Int = 0): DenseMatrix[Double] = {
     val d = A.cols
 
-    val omega = new DenseMatrix(d, l, randn(d*l).toArray) //cpu: d*l, mem: d*l
+    implicit val randBasis: RandBasis = new RandBasis(new ThreadLocalRandomGenerator(new MersenneTwister(seed)))
```
tomerk: any reason why this is marked implicit even though you manually pass it in the next line?

etrain:
Also - it would be a good idea to think about how we want to handle microbenchmarks. I like the idea of running them/tracking them regularly, but maybe they should have separate entry points which would sit outside of the test infrastructure.

Would be easy enough to have Jenkins run them as a separate target and keep the logs around.

Any thoughts?

src/main/scala/nodes/learning/ApproximatePCA.scala
etrain: No, will fix.

shivaram:
LGTM. Yes - we can use tags to mark long running tests. See http://alvinalexander.com/scala/scalatest-mark-tests-tags-to-include-exclude-sbt for an example.

shivaram:
Merging this

PR #237, opened 2016-03-10
Switching norm from singular to frobenius due to conditioning issues.

etrain:
cc @shivaram based on a conversation we had offline.


shivaram:
LGTM - Merging this

PR #247, opened 2016-03-12
Port Cifar Augmented pipeline and its nodes

### LINKED ISSUE #242: Port CIFAR data augmentation to KeystoneML master

shivaram:
Fixes #242


shivaram:
cc @ericmjonas @Vaishaal 

Some notes:
1. This lacks any unit tests right now. I wanted to get this out early to get some feedback on interfaces. I will work on unit tests soon

2. To do augmentation I was forced to use `FunctionNode` -- I don't think we have a better way to do this ? Also not being able to chain function nodes in the middle of a pipeline is a pain.

3. It'll be good to see if some of these APIs like `RandomFlipper` can be generalized like @ericmjonas mentioned in the other thread.

src/main/scala/evaluation/AugmentedExamplesEvaluator.scala
```diff
@@ -0,0 +1,69 @@
+package evaluation
+
+import scala.reflect.ClassTag
+import breeze.linalg._
+import breeze.numerics._
+
+import org.apache.spark.rdd.RDD
+import nodes.util.MaxClassifier
+
+object AugmentedExamplesEvaluator extends Serializable {
```
etrain: Instead of `AugmentedExamplesEvaluator` this seems like it could just as easily be called `EnsembleEvaluator` or something. 

In the traditional ensemble setting, a hypothetical `EnsembleEstimator` could be an `Estimator[T,Seq[V]]` or if people wanted to construct ensembles themselves, they could use a `Pipeline.gather` to coallate the results of running several models on an input data item.

Here we have a case that's slightly different, but I think the basic idea still holds.

src/main/scala/evaluation/AugmentedExamplesEvaluator.scala
```diff
@@ -0,0 +1,69 @@
...
+  object AggregationPolicyType extends Enumeration {
+    type AggregationPolicyType = Value
+    val average, borda = Value
+  }
+
+  def averagePolicy(preds: Array[DenseVector[Double]]): DenseVector[Double] = {
+    preds.reduce(_ + _) :/ preds.size.toDouble
+  }
+
+  // * Let s(k) be the ordering of patch k.
```
etrain: Maybe change this to a scaladoc?

src/main/scala/evaluation/AugmentedExamplesEvaluator.scala
```diff
@@ -0,0 +1,69 @@
...
+  def bordaPolicy(preds: Array[DenseVector[Double]]): DenseVector[Double] = {
+    val ranks = preds.map { vec =>
+      val sortedPreds = vec.toArray.zipWithIndex.sortBy(_._1).map(_._2)
+      val rank = DenseVector(sortedPreds.zipWithIndex.sortBy(_._1).map(x => x._2.toDouble))
+      rank
+    }
+    ranks.reduceLeft(_ + _)
+  }
+
+  def apply[T: ClassTag](
```
etrain: Given my comment above - I wonder if this names RDD is the right thing to do, or whether the grouping to a `Array[DenseVector[Double]]` makes sense in another node? What I'm struggling with a little bit conceptually here is the logical pipeline that we want to describe. Conceptually, featurization is like a map from an Image to a Matrix of features. Those features are concatenated and a model is fit. At test time there is no concatenation and we need an averaging policy. 

This is just one way to look at it, though - and I guess it is worth sketching out variants of the pipeline on a whiteboard or something.

src/main/scala/nodes/images/CenterCornerPatcher.scala
```diff
@@ -0,0 +1,48 @@
...
+  * Extract four corner patches and the center patch of the specified size.
+  * If flips is set to true, then horizontal flips of all 5 patches is also
+  * returned
+  *
+  * @param patchSizeX size of patch along xDim
+  * @param patchSizeY size of patch along yDim
+  * @param horizontalFlips if horizontal flips of patches should also be returned
+  * @return patches of size patchSizeX x patchSizeY
+  */
+case class CenterCornerPatcher(
```
etrain: Can this be accomplished with a correctly configured `Windower`,  `HorizontalFlips`, and a `Gather` - not saying we wouldn't also want a node called this that is the composition of these things - just trying to avoid having a node that feels pretty special purpose.

src/main/scala/nodes/images/Pooler.scala
```diff
@@ -67,3 +67,21 @@ class Pooler(
     ChannelMajorArrayVectorizedImage(patch, ImageMetadata(numPoolsX, numPoolsY, numChannels))
   }
 }
+
+object Pooler {
+
+  def sumPooler(in: DenseVector[Double]): Double = {
```
etrain: What's wrong with `sum(in)`?

src/main/scala/nodes/images/Pooler.scala
```diff
@@ -67,3 +67,21 @@ class Pooler(
...
+    var sum = 0.0
+    while (i < in.length) {
+      sum = sum + in(i)
+      i = i + 1
+    }
+    sum
+  }
+
+  def avgPooler(in: DenseVector[Double]): Double = {
+    sumPooler(in) / in.length
```
etrain: Similarly, why not `breeze.stats.mean`?

src/main/scala/nodes/images/RandomPatcher.scala
```diff
@@ -0,0 +1,47 @@
+package nodes.images
+
+import org.apache.spark.rdd.RDD
+
+import utils.{ImageUtils, Image}
+import pipelines.FunctionNode
+
+/**
+  * Extract random square patches from an image
```
etrain: We should probably mention that patches are selected *uniformly* at random.

etrain:
Alright - code-wise, this mostly looks good. The higher level question is how we want to represent this pipeline conceptually. Is it a flatmap with some key attached to the features that we can later group on (this is messy because it assumes labels and features have identical partitioning, etc.) or would we rather think of this in more general terms with some kind of ensembling. I think this is one of those situations where a whiteboard and an example pipeline might be useful, then we can decide if we have sufficient tools to represent that thing or not.

src/main/scala/nodes/images/Pooler.scala
shivaram: Its slower - and constantly showed up in `jstack` while I was running experiments. To get a sense of how much slower here is a unscientific benchmark I ran on my laptop
```
var sum=0.0; var sumTime=0.0; 
for (i <- 0 until 10000) { 
  val a = DenseVector.rand(4096);
  val begin = System.nanoTime(); 
  sum += sumPooler(a);  // Replace this with a.sum for the Breeze one
  sumTime += (System.nanoTime() - begin)/1000.0 
};
```
Avg sumTime for breeze: 163.9452 microseconds
Avg sumTime for while loop: 21.4573 microseconds

At length=8192
Avg sumTime for breeze: 328.2782 microseconds
Avg sumTime for while loop: 40.1276 microseconds

At length=81920
Avg sumTime for breeze: 3351.0207 microseconds
Avg sumTime for while loop: 412.5542 microseconds

So this 8x overhead seems to be consistent as we increase the vector length.

shivaram:
Cool - We can chat about this on Monday. I don't think we have the right machinery to do this kind of augment-ensembles cleanly yet, but we can try to use this pipeline to design things

src/main/scala/nodes/images/Pooler.scala
etrain: Hmm.. ".sum" is deprecated, and on *my* laptop - `sum(x)` (well, with an appropriate change to "var sum=0.0" above), goes 2x faster than this while loop.

shivaram: I see the same thing - Not sure why the while loop might be slower, but I feel this isn't worth investigating. Changed it to just use `sum(_)` now

src/main/scala/nodes/images/RandomFlipper.scala
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+  * Flips an image horizontally with given probability
+  *
+  * @param flipChance probability that an image should be flipped
+  * @return image flipped image or original image
+  */
+case class RandomFlipper(flipChance: Double, seed: Long = 12334L) extends Transformer[Image, Image] {
+
+  val rnd = new scala.util.Random(seed)
```
Vaishaal: Apply scala.util.Random isn't serializable...

```
Caused by: java.io.NotSerializableException: scala.util.Random
Serialization stack:
        - object not serializable (class: scala.util.Random, value: scala.util.Random@39652a30)
        - field (class: nodes.images.RandomPatcher, name: rnd, type: class scala.util.Random)
        - object (class nodes.images.RandomPatcher, <function1>)
        - field (class: nodes.images.RandomPatcher$$anonfun$apply$1, name: $outer, type: class nodes.images.RandomPatcher)
        - object (class nodes.images.RandomPatcher$$anonfun$apply$1, <function1>)
        at org.apache.spark.serializer.SerializationDebugger$.improveException(SerializationDebugger.scala:40)
        at org.apache.spark.serializer.JavaSerializationStream.writeObject(JavaSerializer.scala:47)
        at org.apache.spark.serializer.JavaSerializerInstance.serialize(JavaSerializer.scala:84)
        at org.apache.spark.util.ClosureCleaner$.ensureSerializable(ClosureCleaner.scala:301)
        ... 22 more
16/03/13 23:33:37 ERROR ErrorMonitor: AssociationError [akka.tcp://sparkDriver@192.168.10.51:48870] <- [ak
```

src/main/scala/nodes/images/CenterCornerPatcher.scala
```diff
@@ -0,0 +1,48 @@
...
+      centerCornerPatchImage(x)
+    }
+  }
+
+  def centerCornerPatchImage(in: Image): Iterator[Image] = {
+    val xDim = in.metadata.xDim
+    val yDim = in.metadata.yDim
+
+    val startXs = Array(0, xDim-patchSizeX, 0, xDim-patchSizeX, (xDim-patchSizeX)/2)
+    val startYs = Array(0, 0, yDim-patchSizeY, (yDim-patchSizeY)/2)
```
Vaishaal: I think these two lists should be the same length

src/main/scala/nodes/images/RandomPatcher.scala
```diff
@@ -0,0 +1,47 @@
...
+  * @param patchSizeY size of each patch along yDim
+  * @return numPatches images of size patchSizeX x patchSizeY
+  */
+case class RandomPatcher(
+    numPatches: Int,
+    patchSizeX: Int,
+    patchSizeY: Int,
+    seed: Long = 12334L) extends FunctionNode[RDD[Image], RDD[Image]] {
+
+  val rnd = new scala.util.Random(seed)
```
Vaishaal: Same deal about serializability...

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
```diff
@@ -0,0 +1,158 @@
...
+
+        ((unnormFilters(::, *) / (twoNorms + 1e-10)) * whitener.whitener.t, whitener)
+    }
+
+    val trainImagesAugmented = RandomFlipper(flipChance).apply(
+      RandomPatcher(conf.numRandomPatchesAugment, augmentPatchSize, augmentPatchSize).apply(
+        trainImages))
+
+    val unscaledFeaturizer = 
+      new Convolver(filters, augmentPatchSize, augmentPatchSize, numChannels, Some(whitener), true)
```
Vaishaal: Do you really want to convolve with 24 x 24 patches?

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
```diff
@@ -0,0 +1,158 @@
...
+        //Normalize them.
+        val sampleFilters = MatrixUtils.sampleRows(baseFilterMat, conf.numFilters)
+        val unnormFilters = whitener(sampleFilters)
+        val unnormSq = pow(unnormFilters, 2.0)
+        val twoNorms = sqrt(sum(unnormSq(*, ::)))
+
+        ((unnormFilters(::, *) / (twoNorms + 1e-10)) * whitener.whitener.t, whitener)
+    }
+
+    val trainImagesAugmented = RandomFlipper(flipChance).apply(
```
Vaishaal: Mostly for my own curiosity..why do we augment train images differently than test?

src/main/scala/nodes/images/RandomFlipper.scala
shivaram: Uh - thats a pain. Using `java.util.Random` now

src/main/scala/nodes/images/CenterCornerPatcher.scala
shivaram: Nice catch. Fixed now

src/main/scala/nodes/images/RandomPatcher.scala
shivaram: Done

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
shivaram: That argument is the imageSize as documented in https://github.com/amplab/keystone/blob/master/src/main/scala/nodes/images/Convolver.scala#L22
The patchSize comes from the filters (hmm we are calling both the 6x6 things and the 24x24 things as patches ?)

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
shivaram: We did this because the original cuda-convnet https://code.google.com/p/cuda-convnet/ code did it this way

src/main/scala/nodes/images/Pooler.scala
shivaram: Removed now

src/main/scala/nodes/images/RandomPatcher.scala
shivaram: Done

src/main/scala/evaluation/AugmentedExamplesEvaluator.scala
shivaram: Done

shivaram:
I've fixed all the comments other the discussion about ensembling. I also introduced a new class `RandomImageTransformer` instead of the `RandomFlipper` per @ericmjonas 's request.

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
Vaishaal: yeah that was a point of confusion for me this weekend.

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
shivaram: FWIW this is the link to the writeup https://code.google.com/p/cuda-convnet/wiki/TrainingNet#Training_on_image_translations -- Significant bits are

> ...This will cause the net to train on random 24x24 patches...
After training is done... compute the test error as averaged over 10 patches of the original images (4 corner patches, center patch, and their horizontal reflections). This will produce better results than testing on the center patch alone

src/main/scala/nodes/images/CenterCornerPatcher.scala
shivaram: Well we need a function node to do this `flatMap`. And I don't think the windower is appropriate here is it gets all the matches with size patchSize (we can throw out the rest of them but that seems like same amount of code with wasted work). 

I think having `ImageUtils.crop` and `ImageUtils.flipHorizontal` make this function pretty readable, but yeah this node does seem overly-specialized

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
```diff
@@ -0,0 +1,158 @@
...
+        //Normalize them.
+        val sampleFilters = MatrixUtils.sampleRows(baseFilterMat, conf.numFilters)
+        val unnormFilters = whitener(sampleFilters)
+        val unnormSq = pow(unnormFilters, 2.0)
+        val twoNorms = sqrt(sum(unnormSq(*, ::)))
+
+        ((unnormFilters(::, *) / (twoNorms + 1e-10)) * whitener.whitener.t, whitener)
+    }
+
+    val trainImagesAugmented = RandomImageTransformer(flipChance, ImageUtils.flipHorizontal).apply(
```
etrain: is this written this way because `FunctionNode` isn't a first-class citizen?

shivaram: Yeah - you can't chain anything after function nodes right now

shivaram:
Alright added unit tests for the two patcher nodes and one for `flipHorizontal`. I'm going to run the entire pipeline on the local cluster to make sure we get the same results we got for the paper. @etrain Let me know if you have any other comments on this.

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
```diff
@@ -0,0 +1,158 @@
...
+
+    // Load up training data, and optionally sample.
+    val trainData = conf.sampleFrac match {
+      case Some(f) => CifarLoader(sc, conf.trainLocation).sample(false, f).cache
+      case None => CifarLoader(sc, conf.trainLocation).cache
+    }
+    val trainImages = ImageExtractor(trainData)
+
+    val patchExtractor = new Windower(conf.patchSteps, conf.patchSize)
+      .andThen(ImageVectorizer.apply)
```
etrain: style nit - can we move the `andThen` to the end of the previous line and remove the spurious `.`, `(`, `)` and `apply` here? I know you're copy/pasting old code, but I like that style better.

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
etrain: Hmm.. I wonder if there's a better word to use for the larger "patch" - something like "subImage" or something?

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
shivaram: Done

shivaram:
Other than the performance problem with repeated convolutions, the test error was fine. I got 
```
16/03/21 01:05:04 INFO RandomPatchCifarAugmented: Test error is: 0.1833
```
from using 5120 filters (i.e. 40k features)

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
```diff
@@ -0,0 +1,157 @@
...
+        ((unnormFilters(::, *) / (twoNorms + 1e-10)) * whitener.whitener.t, whitener)
+    }
+
+    val trainImagesAugmented = RandomImageTransformer(flipChance, ImageUtils.flipHorizontal).apply(
+      RandomPatcher(conf.numRandomPatchesAugment, augmentPatchSize, augmentPatchSize).apply(
+        trainImages))
+
+    val unscaledFeaturizer = 
+      new Convolver(filters, augmentPatchSize, augmentPatchSize, numChannels, Some(whitener), true)
+        .andThen(SymmetricRectifier(alpha=conf.alpha))
```
etrain: The next ten or so lines have the old `.andThen(` syntax still.

src/main/scala/evaluation/AugmentedExamplesEvaluator.scala
shivaram: This is now tracked in #250 and #249

src/main/scala/evaluation/AugmentedExamplesEvaluator.scala
shivaram: Same as above - This is now tracked in #250 and #249

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
shivaram: Fixed all the `andThen`s now

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
shivaram: I've now just called it image - i.e. `augmentedImgSize` and `numRandomImagesAugmented` -- So other than the nodes the only way `patch` is used is to refer to the convolver patches. Let me know if this looks ok

shivaram:
@etrain the latest commit https://github.com/amplab/keystone/pull/247/commits/202317e2ee045f12373c00a18b12541602575b9f updates the other CIFAR / Mnist pipeline to also use the style described in #258 -- Tests pass, but it would still be good to get another pair of eyes on this commit

src/main/scala/pipelines/images/cifar/RandomCifar.scala
```diff
@@ -41,18 +47,10 @@ object RandomCifar extends Serializable with Logging {
...
-
-    val trainFeatures = featurizer(trainImages)
-    val trainLabels = labelExtractor(trainData)
-
-    val model = LinearMapEstimator(conf.lambda).fit(trainFeatures, trainLabels)
+        new Cacher[DenseVector[Double]] andThen
+        (LinearMapEstimator(conf.lambda), trainImages, trainLabels)
 
-    val predictionPipeline = featurizer andThen model andThen MaxClassifier
+    val predictionPipeline = featurizer andThen MaxClassifier
```
etrain: can we add the `MaxClassfier` to the end of what we're calling `featurizer` and change that name to `predictionPipeline`?

src/main/scala/pipelines/images/cifar/RandomPatchCifar.scala
```diff
@@ -50,26 +56,16 @@ object RandomPatchCifar extends Serializable with Logging {
...
-      ClassLabelIndicatorsFromIntLabels(numClasses) andThen
-      new Cacher[DenseVector[Double]]
-
-    val trainFeatures = featurizer(trainImages)
-    val trainLabels = labelExtractor(trainData)
-
-    val model = new BlockLeastSquaresEstimator(4096, 1, conf.lambda.getOrElse(0.0)).fit(trainFeatures, trainLabels)
-
-    val predictionPipeline = featurizer andThen model andThen MaxClassifier andThen new Cacher[Int]
+    val predictionPipeline = featurizer andThen MaxClassifier andThen new Cacher[Int]
```
etrain: Same thing here?

src/main/scala/pipelines/images/cifar/LinearPixels.scala
```diff
@@ -23,23 +23,20 @@ object LinearPixels extends Logging {
...
+    // A featurizer maps input images into vectors. For this pipeline, we'll also convert the image to grayscale.
+    // We estimate our model by calling a linear solver on our data.
+    val featurizer = GrayScaler andThen
+      ImageVectorizer andThen
+      (new LinearMapEstimator, trainImages, trainLabels)
 
-    // The final prediction pipeline is the composition of our featurizer and our model.
-    // Since we end up using the results of the prediction twice, we'll add a caching node.
-    val predictionPipeline = featurizer andThen model andThen MaxClassifier
+    val predictionPipeline = featurizer andThen MaxClassifier
```
etrain: And similarly here.

src/main/scala/pipelines/images/mnist/MnistRandomFFT.scala
```diff
@@ -43,9 +43,10 @@ object MnistRandomFFT extends Serializable with Logging {
...
-    val pipeline = featurizer
-        .andThen(new BlockLeastSquaresEstimator(conf.blockSize, 1, conf.lambda.getOrElse(0)), train.data, labels)
-        .andThen(MaxClassifier)
+    val pipeline = featurizer andThen
+        (new BlockLeastSquaresEstimator(conf.blockSize, 1, conf.lambda.getOrElse(0)),
+          train.data, labels) andThen
+        MaxClassifier
 
     // Train the model
     val model = pipeline
```
etrain: this seems spurious?

shivaram: removed all uses of `model` now

src/main/scala/pipelines/images/cifar/RandomPatchCifar.scala
shivaram: done

src/main/scala/pipelines/images/cifar/RandomCifar.scala
shivaram: Yep - done

src/main/scala/pipelines/images/cifar/LinearPixels.scala
shivaram: Done

etrain:
This LGTM, merging. Thanks @shivaram !

Issue #253, opened 2016-03-18
Document Release Procedure


etrain:
Will be useful for cutting v0.3. In particular, we should update RELEASE.md

shivaram:
Closed by #254

PR #254, opened 2016-03-18
Updated the KeystoneML release guide

tomerk:
@etrain @shivaram

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -38,11 +38,22 @@ resolvers ++= Seq(
 
 ## Full Release
 
-1. Update the build.sbt and the Makefile to the new version
+1. Update the build.sbt and the Makefile to reference the new version
 
 2. Cut a release branch for major version releases
+```
+git checkout -b branch-v[major_version_here]
+git add build.sbt
+git add Makefile
+git commit -m "Cut release branch for [major_version_here]"
+```
 
-3. Tag the specific commit of the release branch to be released with its version
+3. Tag the specific commit of the release branch to be released with its version, and push to the upstream AMPLab
+KeystoneML repo.
+```
+git tag v[full_version_here]
+git push [upstream_keystoneml_reference] branch-v[major_version_here] --tags
+```
 
 4. We publish KeystoneML releases and snapshots to sonatype repository https://oss.sonatype.org/. To
 create a release you will first need to create a SonaType account.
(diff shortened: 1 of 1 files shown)


tomerk:
Fixes issue #253

RELEASE.md
```diff
@@ -34,4 +36,39 @@ resolvers ++= Seq(
   )
 ```
 
-## TODO(tomerk): Add details about publishing a full release
+## Full Release
+
+1. Update build.sbt to the correct version
```
etrain: And Makefile!

etrain:
Having never done this before, this LGTM. Do you have any further comments @shivaram ?

RELEASE.md
```diff
@@ -34,4 +36,39 @@ resolvers ++= Seq(
   )
 ```
 
-## TODO(tomerk): Add details about publishing a full release
+## Full Release
+
+1. Update the build.sbt and the Makefile to the new version
+
+2. Cut a release branch for major version releases
```
shivaram: Can you be explicit about the git commands to run for this and the next step ?

RELEASE.md
```diff
@@ -34,4 +36,39 @@ resolvers ++= Seq(
...
+
+7. Add your Sonatype credentials to `~/.sbt/0.13/sonatype.sbt`
+credentials += Credentials("Sonatype Nexus Repository Manager",
+                           "oss.sonatype.org",
+                           "<userame>",
+                           "<password>")
+
+8. Run `sbt/sbt publishSigned` to upload to sonatype a staging repository.
+
+9. Follow [these instructions](http://central.sonatype.org/pages/releasing-the-deployment.html) to
```
shivaram: According to https://github.com/xerial/sbt-sonatype#publishing-your-artifact we can just do `sbt sonatypeRelease`. Does that work for keystone ?

RELEASE.md
tomerk: I can be explicit about the next step (tagging the commit). This step is much more variable depending on what exact commits we want to include (whether we want to do any commit cherry-picking, etc.)

shivaram: Write it from the perspective of cutting a brand new release from the `master` branch. We can add a subsection below for incremental releases. 

Also almost all the releases should happen from a branch (say `master` or `branch-0.2`) and shouldn't require cherry-picking. We might cherrypick before as a part of backporting things.

RELEASE.md
tomerk: Hmm I wasn't aware, I've never tried it. It might work?

shivaram: Can we make a note to try this for the 0.3 release and see if it works ? It'll be cool to not have to mess with the web UI for just the promoting bits.

tomerk:
Okay @etrain @shivaram I haven't explicitly tested these git commands, but I they should roughly work.

shivaram:
Ok thanks. LGTM. Lets merge this and if during 0.3 we find something that doesn't work we can come back and update the doc.

Issue #187, opened 2015-12-09
Auto-Selected Least Squares Solver


tomerk:
Create a node / optimizer combo somehow to specify that a least squares solver is desired, and have the Optimizer automatically choose the right solver.

shivaram:
Fixed by #256

PR #256, opened 2016-03-20
Added the optimizable least squares estimator

tomerk:
@etrain @shivaram


src/main/scala/nodes/learning/LeastSquaresEstimator.scala
```diff
@@ -0,0 +1,87 @@
...
+      (solver, Sparsify() andThen (solver, _, _))
+    },
+    {
+      val solver = new BlockLeastSquaresEstimator(1000, 3, lambda = lambda)
+      (solver, Densify() andThen (solver, _, _))
+    },
+    {
+      val solver = new LinearMapEstimator(Some(lambda))
+      (solver, Densify() andThen (solver, _, _))
+    }
```
shivaram: Is there a specific reason why `LocalLeastSquaresEstimator` isn't in this list ?

src/main/scala/nodes/learning/LBFGS.scala
```diff
@@ -183,8 +208,13 @@ class SparseLBFGSwithL2(
     val numCorrections: Int = 10,
     val convergenceTol: Double = 1e-4,
     val numIterations: Int = 100,
-    val regParam: Double = 0.0)
-  extends LabelEstimator[SparseVector[Double], DenseVector[Double], DenseVector[Double]] {
+    val regParam: Double = 0.0,
+    val sparseOverhead: Double = 8)
```
shivaram: Can you document what `sparseOverhead` is ? (Also I'm genuinely curious about what it is ?)

src/main/scala/nodes/learning/LeastSquaresEstimator.scala
```diff
@@ -0,0 +1,87 @@
...
+
+import scala.reflect._
+
+/**
+ * A least squares solver that is optimized to use a fast algorithm, based on characteristics of the
+ * workload and cost models.
+ *
+ * Currently selects between Dense LBFGS, Sparse LBFGS, Exact Distributed Solve, and Approximate Block Solve.
+ *
+ * The default weights were determined empirically via results run on a 16 r3.4xlarge node cluster.
```
shivaram: So this is a bit unrelated to this PR but it would be good to check in a weights estimator script in `scripts/` or something like that.

src/main/scala/nodes/learning/LeastSquaresEstimator.scala
tomerk: When @etrain made the cost models for me I don't think he bothered w/ this one because it's about equal to DenseLBFGS on the VOC workload

src/main/scala/nodes/learning/LBFGS.scala
tomerk: It's a constant @etrain and I are using for roughly estimating how much slower an operation on fully filled sparse data is versus the equivalent data stored in dense form. I'll add a comment.

tomerk:
Will close issue #187

src/test/scala/nodes/learning/LeastSquaresEstimatorSuite.scala
```diff
@@ -0,0 +1,92 @@
+package nodes.learning
+
+import breeze.linalg.{DenseVector, SparseVector}
+import org.apache.spark.SparkContext
+import org.scalatest.FunSuite
+import pipelines._
+import workflow.WorkflowUtils
+
+class LeastSquaresEstimatorSuite extends FunSuite with LocalSparkContext with Logging {
```
shivaram: Very nice unit tests ! Minor question: 

1. If you set `k = 2` as in binary classification does DenseLBFGS get chosen over the BlockSolver ?
2. Does this test take very long to run ? Just curious if it'll be fine to put in Jenkins etc.

src/main/scala/nodes/learning/LeastSquaresEstimator.scala
tomerk: Bringing @etrain into the loop since he was in charge of building the cost models

shivaram:
Thanks @tomerk -- This is a very nice change. I just had a few inline comments.

src/main/scala/nodes/learning/LeastSquaresEstimator.scala
shivaram: Hmm I see. Can we just add the cost model to the file and let the optimizer figure out when it'll be useful ? It should be a simple cost model. The cost model is something like 
```
 Communication = n*d + n*k   // we collect both matrices
 Computation (CPU) = n*d*n + 2/3*n^3 + d*n*k
 // The first is for computing A*A.t, 
 // the second for inverting the matrix, 
 // the last is for multiply with A.t
```

src/main/scala/nodes/learning/LeastSquaresEstimator.scala
etrain: Here's a gist with the code I used: https://gist.github.com/etrain/7486bdf1e7b29e381a17

And the data: https://gist.github.com/etrain/f9b05ea7e882f4740063

One important thing about this is instead of doing 
`totalTime = max(k1*cpuTime, k2*memTime) + k3*networkTime`
it estimates
`totalTime = k1*cpuTime + k2*memTime + k3*networkTime` 

But in most cases (e.g. memory and cpu are not well balanced) these things should be functionally equivalent.
Also - I'm a little worried that minimizing least squares loss biases us toward fitting constants in regimes that take a long time, but again, this seems fine and the important bit - that this seems to make good predictions - is maintained.

src/main/scala/nodes/learning/LeastSquaresEstimator.scala
etrain: I'm all for including this if it's simple. Do we have any empirical data we can verify this cost model on? Basically I want to make sure that fitting this matches our intuition.

src/main/scala/nodes/util/Densify.scala
```diff
@@ -0,0 +1,21 @@
+package nodes.util
+
+import breeze.linalg.DenseVector
+import breeze.linalg.Vector
+import workflow.Transformer
+
+/**
+ * Transformer to densify vectors into DenseVectors.
+ */
+case class Densify[T <: Vector[Double]]() extends Transformer[T, DenseVector[Double]] {
```
etrain: I think this is a fine solution - basically makes this a no-op if it can be, and densify it if necessary.

etrain:
Nice stuff @tomerk ! This LGTM - feel free to just stick the scripts used to estimate the constants in the "scripts/" directory.

src/main/scala/nodes/learning/LeastSquaresEstimator.scala
shivaram: The VOC data would be the best to verify this on - Did we use this solver for any of our SIGMOD expts ? If not we'll need to run a separate experiment for this

src/main/scala/nodes/learning/LeastSquaresEstimator.scala
shivaram: Ah I see. I didn't realize we fit a linear model from experiments to get the constants. I somehow assumed we did this by running some microbenchmarks that say used raw network / mem-bw etc. This is more like the Planck / Ernest. I'm fine with pulling this R script into `scripts/` for now. We can return to setup a more automated way to do this in the future

etrain: Yeah, in the current setup the constants that are estimated can be mapped back out to GFLOP/s, memory GB/s, and network gbps - i've tried it both ways and got similar results, but this is where the final constants here come from.

src/test/scala/nodes/learning/LeastSquaresEstimatorSuite.scala
tomerk: 1. Doesn't look like it in the existing "big n big d dense" test case I have, should it?
2. No, it takes seconds.

src/main/scala/nodes/learning/LeastSquaresEstimator.scala
tomerk: We didn't stress test this solver at different data settings like we did the other solvers. We did use it for the VOC experiment though.

src/main/scala/nodes/learning/LBFGS.scala
tomerk: Done

tomerk:
Besides Local Least Squares stuff, I've taken care of all the comments now

Issue #239, opened 2016-03-11
Optimizable Nodes


etrain:
We have a few nodes that should be `Optimizable` - Fisher Vector, GMM, PCA. Useful for the VOC Pipeline.

tomerk:
Removing from milestone 0.3.0. Delaying until we get a chance to set up proper PCA cost modeling.

etrain:
Closed by #257

PR #257, opened 2016-03-20
Added optimizable gmm fisher vector, column pca estimators

tomerk:
issue #239

### DIFF
--- a/src/main/scala/nodes/images/FisherVector.scala
+++ b/src/main/scala/nodes/images/FisherVector.scala
@@ -62,7 +62,7 @@ case class FisherVector(gmm: GaussianMixtureModel)
  *
  * @param k Number of centers to estimate.
  */
-case class GMMFisherVectorEstimator(k: Int) extends Estimator[DenseMatrix[Float], DenseMatrix[Float]] {
+case class ScalaGMMFisherVectorEstimator(k: Int) extends Estimator[DenseMatrix[Float], DenseMatrix[Float]] {
   protected def fit(data: RDD[DenseMatrix[Float]]): FisherVector = {
     val gmmTrainingData = data.flatMap(x => convert(MatrixUtils.matrixToColArray(x), Double))
     val gmmEst = new GaussianMixtureModelEstimator(k)
@@ -81,15 +81,15 @@ case class GMMFisherVectorEstimator(k: Int) extends Estimator[DenseMatrix[Float]
  * @param k Number of centers to estimate.
  */
 
-case class OptimizableGMMFisherVectorEstimator(k: Int) extends OptimizableEstimator[DenseMatrix[Float], DenseMatrix[Float]] {
-  val default = GMMFisherVectorEstimator(k)
+case class GMMFisherVectorEstimator(k: Int) extends OptimizableEstimator[DenseMatrix[Float], DenseMatrix[Float]] {
+  val default = ScalaGMMFisherVectorEstimator(k)
 
   override def optimize(sample: RDD[DenseMatrix[Float]], numPerPartition: Map[Int, Int])
   : RDD[DenseMatrix[Float]] => Pipeline[DenseMatrix[Float], DenseMatrix[Float]] = {
     if (k >= 32) {
-      nodes.images.external.GMMFisherVectorEstimator(k).withData(_)
+      nodes.images.external.EncEvalGMMFisherVectorEstimator(k).withData(_)
     } else {
-      GMMFisherVectorEstimator(k).withData(_)
+      ScalaGMMFisherVectorEstimator(k).withData(_)
     }
   }
 }
--- a/src/main/scala/nodes/images/FisherVector.scala
+++ b/src/main/scala/nodes/images/FisherVector.scala
@@ -3,8 +3,10 @@ package nodes.images
 import breeze.linalg._
 import breeze.numerics._
 import breeze.stats.mean
-import nodes.learning.GaussianMixtureModel
-import workflow.Transformer
+import nodes.learning.{GaussianMixtureModelEstimator, GaussianMixtureModel}
+import org.apache.spark.rdd.RDD
+import utils.MatrixUtils
+import workflow.{Pipeline, OptimizableEstimator, Estimator, Transformer}
 
 /**
  * Abstract interface for Fisher Vector.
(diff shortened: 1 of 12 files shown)


src/main/scala/nodes/images/FisherVector.scala
```diff
@@ -49,3 +51,45 @@ case class FisherVector(gmm: GaussianMixtureModel)
...
+ * Trains either a scala or an `enceval` Fisher Vector implementation, via
+ * estimating a GMM by treating each column of the inputs as a separate
+ * DenseVector input to [[GaussianMixtureModelEstimator]]
+ *
+ * Automatically decides which implementation to use when node-level optimization is enabled.
+ *
+ * @param k Number of centers to estimate.
+ */
+
+case class OptimizableGMMFisherVectorEstimator(k: Int) extends OptimizableEstimator[DenseMatrix[Float], DenseMatrix[Float]] {
```
etrain: Can we not call these things `Optimizable` when they are? I really don't want users to have to think about what's going on under the covers - the fact that the system is optimizing this under the covers should be incidental to their experience.

etrain: Even if this means changing the name of the old one!

src/main/scala/nodes/learning/PCA.scala
```diff
@@ -55,6 +56,44 @@ case class ColumnPCAEstimator(dims: Int) extends Estimator[DenseMatrix[Float], D
...
+/**
+ * Estimates a PCA model for dimensionality reduction based on a sample of a larger input dataset.
+ * Treats each column of the input matrices like a separate DenseVector input to [[PCAEstimator]] or
+ * [[DistributedPCAEstimator]].
+ *
+ * Automatically decides between distributed and local implementations when node-level optimization is enabled.
+ *
+ * @param dims Dimensions to reduce input dataset to.
+ */
+class OptimizableColumnPCAEstimator(dims: Int) extends OptimizableEstimator[DenseMatrix[Float], DenseMatrix[Float]] {
```
etrain: Same thing here.

src/main/scala/nodes/learning/PCA.scala
```diff
@@ -55,6 +56,44 @@ case class ColumnPCAEstimator(dims: Int) extends Estimator[DenseMatrix[Float], D
...
+  val default = new DistributedColumnPCAEstimator(dims)
+
+  override def optimize(sample: RDD[DenseMatrix[Float]], numPerPartition: Map[Int, Int])
+  : RDD[DenseMatrix[Float]] => Pipeline[DenseMatrix[Float], DenseMatrix[Float]] = {
+    val numFeatures = sample.first().rows
+
+    if (numFeatures <= 100000) {
```
etrain: I don't think this should be "numFeatures" - should instead be `sample.count()`

tomerk:
Okay @etrain @shivaram

src/main/scala/nodes/learning/PCA.scala
```diff
@@ -47,11 +48,111 @@ case class BatchPCATransformer(pcaMat: DenseMatrix[Float]) extends Transformer[D
...
+
+/**
+ * Estimates a PCA model for dimensionality reduction based on a sample of a larger input dataset,
+ * using a distributed PCA algorithm.
+ * Treats each column of the input matrices like a separate DenseVector input to [[DistributedPCAEstimator]].
+ *
+ * @param dims Dimensions to reduce input dataset to.
+ */
+case class DistributedColumnPCAEstimator(dims: Int) extends Estimator[DenseMatrix[Float], DenseMatrix[Float]]
+  with SolverCostModel {
```
shivaram: Why are all these called `SolverCostModel` ?

src/main/scala/pipelines/images/imagenet/ImageNetSiftLcsFV.scala
```diff
@@ -48,7 +48,7 @@ object ImageNetSiftLcsFV extends Serializable with Logging {
         new BatchPCATransformer(convert(csvread(new File(fname)), Float).t)
       case None =>
         val pca = sampledColumns andThen
-            (ColumnPCAEstimator(numPCADesc), trainingData)
```
shivaram: Any reason we don't want to use the optimized nodes :) ?

src/main/scala/nodes/learning/PCA.scala
```diff
@@ -47,11 +48,111 @@ case class BatchPCATransformer(pcaMat: DenseMatrix[Float]) extends Transformer[D
...
+    numMachines: Option[Int] = None,
+    cpuWeight: Double = 3.8e-4,
+    memWeight: Double = 2.9e-1,
+    networkWeight: Double = 1.32)
+  extends OptimizableEstimator[DenseMatrix[Float], DenseMatrix[Float]] {
+  val localEstimator = new LocalColumnPCAEstimator(dims)
+  val distributedEstimator = new DistributedColumnPCAEstimator(dims)
+  val default = distributedEstimator
+
+  override def optimize(sample: RDD[DenseMatrix[Float]], numPerPartition: Map[Int, Int])
```
shivaram: Can we add a unit test for this similar to the solvers one ?

src/main/scala/nodes/learning/PCA.scala
tomerk: I reused the same trait I used for the linear solvers. Do you have a name you prefer?

src/main/scala/pipelines/images/imagenet/ImageNetSiftLcsFV.scala
tomerk: uh not particularly, I just haven't tested these w/ the optimized nodes yet and am trying to minimize risks. I guess the integration tests will try this out for us?

src/main/scala/nodes/learning/PCA.scala
tomerk: Sure

src/main/scala/nodes/learning/PCA.scala
shivaram: Just `CostModel` would suffice -- Also its an internal API so we can augment / change this in the future I guess

src/main/scala/pipelines/images/imagenet/ImageNetSiftLcsFV.scala
shivaram: Yeah testing should help - More so if we look at these pipelines as examples of how to write pipelines we want to encourage use of the optimizable nodes ? cc @etrain

tomerk:
Unit tests added.

shivaram:
Thanks -- changes look fine. Lets wait for @etrain to also take a pass

src/main/scala/nodes/learning/PCA.scala
```diff
@@ -75,20 +107,50 @@ case class DistributedColumnPCAEstimator(dims: Int) extends Estimator[DenseMatri
...
+ * @param cpuWeight
+ * @param memWeight
+ * @param networkWeight
  */
-class OptimizableColumnPCAEstimator(dims: Int) extends OptimizableEstimator[DenseMatrix[Float], DenseMatrix[Float]] {
-  val default = new DistributedColumnPCAEstimator(dims)
+class ColumnPCAEstimator(
+    dims: Int,
+    numMachines: Option[Int] = None,
+    cpuWeight: Double = 3.8e-4,
```
etrain: We should have a philosophical discussion about whether we want to expose details of optimization at this level of user API, but this is fine for now.

src/main/scala/pipelines/images/imagenet/ImageNetSiftLcsFV.scala
etrain: Yeah, let's just go ahead and do it - in general I think we want to move toward the higher level expression of these pipelines, which includes a bunch of benefits like showing up in integration tests (maybe) and it encourages us to eat our own dog food and really use the optimizer.

etrain:
Would be great to get the PCAEstimator and GMMEstimators in the existing pipelines using the `Optimizable` versions - beyond that this LGTM

tomerk:
@etrain @shivaram I've added the optimizable versions to the existing pipelines

etrain:
LGTM - thanks @tomerk !!

PR #261, opened 2016-03-24
Bumping default spark and hadoop versions.

### LINKED ISSUE #218: Clean up build script
The build script is getting a little crusty, we should:

- Bump spark to 1.6.0
- Bump breeze to 0.12

Not sure what else. We may want to use this as an opportunity to test out the work being done in #189 , since both are pretty big changes and could affect accuracy, etc. of the example pipelines.

etrain:
Closes #218 

cc @shivaram

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -31,7 +31,7 @@ libraryDependencies ++= Seq(
 )
 
 {
-  val defaultSparkVersion = "1.3.1"
+  val defaultSparkVersion = "1.5.2"
   val sparkVersion =
     scala.util.Properties.envOrElse("SPARK_VERSION", defaultSparkVersion)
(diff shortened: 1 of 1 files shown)


shivaram:
LGTM. Just to check

1. Does the breeze version need to be bumped ?
2. Does the Makefile need any changes as described in #218 ?

shivaram:
Also at some point we should switch to Scala 2.11 but lets do that sometime later.

etrain:
I haven't tested with the new breeze so I want to leave it out of this release.

We haven't yet bumped the Keystone versions - I want to get this PR merged, branch and create a branch-v0.3 off current master, in that branch i will change version numbers to be correct in both build.sbt and Makefile (per #218), and then we will advance the version numbers in master. Sound sane?

shivaram:
Yep sounds good.

PR #264, opened 2016-03-29
Added a model selector node

tomerk:
@etrain


tomerk:
Moving to research keystone-tuning repo

PR #265, opened 2016-03-30
Allowed DAG branches to be taken lazily

tomerk:
@etrain @shivaram 
This is key for hyperparameter tuning, checkpointing, lazy block solvers, etc.


etrain:
Can you think of a good way to test these changes for surprising behavior? Either via a unit test or a few end-to-end things?

tomerk:
I could write a unit test that manually constructs a DAG and executes it, making sure it doesn't execute stuff that weren't requested yet (to ensure lazyness)

etrain:
LGTM - merging this. We are going to want to run the integration tests on this before too long.

Issue #266, opened 2016-04-05
AutoCache greedy selection slow on complex pipelines


tomerk:
The autocache greedy cache selection process is insanely slow (> hours) for this complex hyperparameter tuning newsgroups pipeline:

```scala
    val evaluator: (RDD[Int], RDD[Int]) => Double = MulticlassClassifierEvaluator(_, _, numClasses).totalAccuracy

    val predictors: Seq[Pipeline[String, Int]] = for (
      ngrams <- Seq(1, 2);
      tf <- Seq(0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0).map(x => TermFrequency[Seq[String]](y => x));
      reg <- Seq(0.0, 0.1, 1.0, 0.01, 10.0)
    ) yield {
      Trim andThen
        LowerCase() andThen
        Tokenizer() andThen
        NGramsFeaturizer(1 to ngrams) andThen
        tf andThen
        (CommonSparseFeatures(conf.commonFeatures), trainData.data) andThen
        (NaiveBayesEstimator(numClasses, lambda = reg), trainData.data, trainData.labels) andThen
        MaxClassifier
    }

    val predictor = Pipeline.tune(
      predictors,
      trainData.data,
      trainData.labels,
      evaluator
    )
```

The profiling didn't take unreasonably long, it's the actual cache selection and insertion that is taking forever.

etrain:
This was closed by #267

PR #267, opened 2016-04-06
Made autocaching much faster for complex DAGs

tomerk:
fix to issue #266

### DIFF
--- a/src/main/scala/workflow/AutoCacheRule.scala
+++ b/src/main/scala/workflow/AutoCacheRule.scala
@@ -43,12 +43,17 @@ class AutoCacheRule(cachingMode: CachingStrategy) extends Rule with Logging {
    * Get an estimate for how many times each instruction will be executed, assuming
    * the given set of instructions have their outputs cached
    */
-  def getRuns(instructions: Seq[Instruction], cache: Set[Int], nodeWeights: Map[Int, Int]): Map[Int, Int] = {
-    val immediateChildrenByInstruction = getImmediateChildrenByInstruction(instructions)
-
-    instructions.indices.foldRight(Map[Int, Int]()) { case (i, runsByIndex) =>
+  def getRuns(
+      instructions: Seq[Instruction],
+      immediateChildrenByInstruction: Map[Int, Seq[Int]],
+      cache: Set[Int],
+      nodeWeights: Map[Int, Int]
+    ): Array[Int] = {
+    val runsByIndex = new Array[Int](instructions.size)
+    var i = instructions.size - 1
+    while (i >= 0) {
       if (immediateChildrenByInstruction(i).isEmpty) {
-        runsByIndex + (i -> 1)
+        runsByIndex(i) = 1
       }
       else {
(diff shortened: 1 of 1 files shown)


tomerk:
Goes down from > hours to ~8 seconds for a complicated, 784 node DAG

src/main/scala/workflow/AutoCacheRule.scala
```diff
@@ -43,12 +43,17 @@ class AutoCacheRule(cachingMode: CachingStrategy) extends Rule with Logging {
    * Get an estimate for how many times each instruction will be executed, assuming
    * the given set of instructions have their outputs cached
    */
-  def getRuns(instructions: Seq[Instruction], cache: Set[Int], nodeWeights: Map[Int, Int]): Map[Int, Int] = {
-    val immediateChildrenByInstruction = getImmediateChildrenByInstruction(instructions)
```
etrain: So it looks like pre-computing this value ahead of time was the main source of savings. Is that right? Out of curiousity did you find this by profiling?

src/main/scala/workflow/AutoCacheRule.scala
```diff
@@ -423,19 +434,20 @@ class AutoCacheRule(cachingMode: CachingStrategy) extends Rule with Logging {
    * Returns true iff there is still an uncached node whose output is used > once, that would fit in memory
    * if cached
    */
-  def stillRoom(caches: Set[Int], runs: Map[Int, Int], profiles: Map[Int, Profile], spaceLeft: Long): Boolean = {
-    runs.exists { i =>
-      (i._2 > 1) &&
-        (!caches.contains(i._1)) &&
-        (profiles.getOrElse(i._1, Profile(0, 0, 0)).rddMem < spaceLeft)
+  def stillRoom(caches: Set[Int], runs: Array[Int], profiles: Map[Int, Profile], spaceLeft: Long): Boolean = {
```
etrain: This seems better from a stylistic perspective - did the change from `Map` to `Int` buy you anything in terms of performance?

etrain:
Thanks @tomerk ! This LGTM - would be great to have your thoughts on my comments but otherwise I am happy to merge.

src/main/scala/workflow/AutoCacheRule.scala
tomerk: I narrowed it down using timings, and it was taking 4 seconds per time it was executed. `getRuns` ends up getting executed thousands of times.

src/main/scala/workflow/AutoCacheRule.scala
tomerk: It didn't buy much if anything in terms of performance, but I was changing `getRuns` anyway, so I figured I may as well change that too just in case.

src/main/scala/workflow/AutoCacheRule.scala
etrain: Nice catch - thanks for digging into this!

PR #270, opened 2016-04-27
Internal dag update

tomerk:
@etrain @shivaram 
See the google doc I linked in the keystone-dev slack channel for extra context.

Over 63% of this PR is unit tests, and documentation is much of the remainder.

Some feedback about naming I'd like is:
What's a better package name than `internal`, or is that a good enough package name?
Also, what's a better alternative name for `Output`? It's a little weird to me to be passing around `Seq[Output]` as the inputs to `Operators` and their methods.


etrain:
Will you make the google doc public and post a link here! Thanks!

tomerk:
https://docs.google.com/document/d/1Px7bwqxS6OA_vaB4yjzmhrxXaLErxa5qMmD80CwEXLg/edit?usp=sharing

src/main/scala/internal/AnalysisUtils.scala
```diff
@@ -0,0 +1,118 @@
...
+   * @param graph  The graph to use
+   * @return  A topologically sorted ordering of all sources/sinks/nodes in the graph
+   */
+  def linearize(graph: Graph): Seq[GraphId] = {
+    // Sort sinks by id to ensure a deterministic final ordering
+    val sortedSinks = graph.sinks.toSeq.sortBy(_.id)
+
+    sortedSinks.foldLeft(Seq[GraphId]()) {
+      case (linearization, sink) => {
+        linearization ++ linearize(graph, sink).filter(id => !linearization.contains(id)) :+ sink
```
etrain: This logic (in particular the `!linearization.contains` is a little bit tricky to follow. The point is that you don't want to repeat nodes that occur in earlier linearizations, right? Myaybe just a small comment here and above.

src/main/scala/internal/AnalysisUtils.scala
```diff
@@ -0,0 +1,118 @@
...
+   * i.e. All ids with a direct dependency on this source/sink/node.
+   *
+   * @param graph The graph to use
+   * @param id The node/sink/source in the graph
+   * @return The set of all children of that id
+   */
+  def getChildren(graph: Graph, id: GraphId): Set[GraphId] = {
+    id match {
+      case id: NodeOrSourceId => {
+        val childrenNodes = graph.dependencies.filter(_._2.contains(id)).keySet
```
etrain: I don't see this becoming a problem in the short term but we might want to point out that this is `O(VE)` and that it could be done faster with explicit bookkeeping in the `Graph` class pointing in both directions.

src/test/scala/internal/AnalysisUtilsSuite.scala
```diff
@@ -0,0 +1,411 @@
+package internal
```
etrain: Thank you for writing such incredibly thorough tests!

src/main/scala/internal/Graph.scala
```diff
@@ -0,0 +1,425 @@
...
+    sources: Set[SourceId],
+    sinkDependencies: Map[SinkId, NodeOrSourceId],
+    operators: Map[NodeId, Operator],
+    dependencies: Map[NodeId, Seq[NodeOrSourceId]]
+  ) {
+
+  /**
+   * Get the set of ids of all nodes in this graph.
+   */
+  def nodes: Set[NodeId] = operators.keySet
```
etrain: We don't want this to be a val? (same for sinks)

src/main/scala/internal/Graph.scala
```diff
@@ -0,0 +1,425 @@
...
+  /**
+   * Get the operator at a given node in this graph.
+   */
+  def getOperator(id: NodeId): Operator = operators(id)
+
+  /**
+   * Get {@param num} [[NodeId]]s that don't clash with those of existing nodes.
+   */
+  private def nextNodeIds(num: Int): Seq[NodeId] = {
+    val maxId = (nodes.map(_.id) + 0).max
```
etrain: what's with the `+0`? Is that to handle the empty graph?

src/main/scala/internal/Graph.scala
```diff
@@ -0,0 +1,425 @@
...
+   */
+  private def nextSourceId(): SourceId = nextSourceIds(1).head
+
+  /**
+   * Get a [[SinkId]] that doesn't clash with any existing sink in this graph.
+   */
+  private def nextSinkId(): SinkId = nextSinkIds(1).head
+
+  /**
+   * Immutably add a node to this graph.
```
etrain: 👍

src/main/scala/internal/Graph.scala
```diff
@@ -0,0 +1,425 @@
...
+   * @return A pair containing the new graph and the id assigned to the new source
+   */
+  def addSource(): (Graph, SourceId) = {
+    val id = nextSourceId()
+    val newSources = sources + id
+    (copy(sources = newSources), id)
+  }
+
+  /**
+   * Immutably update the dependencies of a node in this graph.
```
etrain: "Immutably update" is kind of an oxymoron - maybe there's a better way to phrase this?

src/main/scala/internal/Graph.scala
```diff
@@ -0,0 +1,425 @@
...
+
+  /**
+   * Immutably remove a source from this graph.
+   * Note: This may leave invalid dangling dependencies on the deleted source,
+   * that must be manually taken care of.
+   *
+   * @param source The id of the source to remove
+   * @return The new graph
+   */
+  def removeSource(source: SourceId): Graph = {
```
etrain: Seems like these remove methods can all be put into a single method call maybe?

src/main/scala/internal/Graph.scala
```diff
@@ -0,0 +1,425 @@
...
+  }
+
+  /**
+   * Immutably replace all dependencies on a given node or source with a new target.
+   *
+   * @param oldDep The id of the old node or source that is the existing dependency target
+   * @param newDep The id of the node or source intended to be the new target
+   * @return The new graph
+   */
+  def replaceDependency(oldDep: NodeOrSourceId, newDep: NodeOrSourceId): Graph = {
```
etrain: If there's one place in this PR I'm concerned about developer ergonomics it's here where the caller has to keep track of id spaes (well, here and `connectGraph`)

src/main/scala/internal/Graph.scala
```diff
@@ -0,0 +1,425 @@
...
+    (newGraph, otherSourceIdMap -- spliceMap.keySet, otherSinkIdMap)
+  }
+
+  /**
+   * Replace a set of nodes and their connections by inserting a different graph in their
+   * place, and specifying how to connect it to the existing graph.
+   *
+   * @param nodesToRemove  The set of nodes to remove
+   * @param replacement  The graph to insert in their place
+   * @param replacementSourceSplice  A specification of how to connect the replacement graph to the existing graph.
```
etrain: Keeping track of/knowing all these connections is hard - but if you assume only the pipeline builder and optimizer APIs will ever call this then we're probably OK.

etrain:
All told this is a pretty great first cut at rewriting the internals and I think should set a nice foundation for optimization, etc.

Perhaps @shivaram can take a brief look at this  with a copy of the doc above open and see if he agrees with the direction.

etrain:
Re: naming - is `Expression` reasonable? Also, `internal` is no good, maybe `workflow.graph` for now?

etrain:
And we should probably mark the whole thing as package protected or something :)

src/main/scala/internal/AnalysisUtils.scala
tomerk: yeah we can always add this in the future if it becomes a problem :)

src/main/scala/internal/AnalysisUtils.scala
tomerk: Sure, I'll add a comment!

src/main/scala/internal/Graph.scala
tomerk: Whoops, will update

src/main/scala/internal/AnalysisUtils.scala
etrain: Maybe a comment inline while its fresh in our minds?

src/main/scala/internal/Graph.scala
tomerk: Yeah, `.max` on an empty set or seq returns an error. I'll add a comment.

tomerk: Actually I'll just add an if statement instead, that'll be more straightforward.

src/main/scala/internal/Graph.scala
tomerk: "Return a copy of this graph with the dependencies of a node updated"?

src/main/scala/internal/Graph.scala
tomerk: So just a single `removeVertex(id: GraphId)` method?

src/main/scala/internal/Graph.scala
etrain: Yeah - same thing wherever you say "Immutably add" or "Immutably update"

src/main/scala/internal/Graph.scala
etrain: That's what I was thinking - though an argument for API symmetry could be made. I have a preference for smaller API surface area over symmetry, but your call if you think programming against `removeNode` is weird?

On another note - I haven't been looking hard for consistency of usage of `Node` vs. `Vertex`, etc. (though I think the only other place `vertex` is used is in the comments and I guess it should be updated?) - In general maybe take a pass looking for common sources of confusion and make sure that we're consistent with language here.

tomerk: I used vertex twice in the initial `Graph` doc to mean (Node or Source or Sink), and consistently using Node to only refer to a vertex with an operator attached. This is probably bad to use vertex in a different way than node, but it's why I don't want to have a single method named `removeNode` to remove nodes, sinks, and sources.

src/main/scala/internal/Graph.scala
tomerk: `replaceDependency` actually doesn't affect id space tracking, because it isn't assigning new ids to existing nodes or sources. It's just moving dependencies around between existing nodes w/o modifying any ids.

Only `addGraph` and `connectGraph` add id space bookkeeping (and I've made sure to isolate the bookkeeping to sources and sinks, instead of having to do node id bookkeeping as well).

src/main/scala/internal/Graph.scala
tomerk: Yeah, the assumption is users never touch any of the graph methods directly, and the only method that will need to use this will have enough contextual knowledge to know what connections they want to make.

tomerk:
Okay, I named it workflow.graph for now. And sure, I could try renaming Output to `Expression`.

src/main/scala/internal/AnalysisUtils.scala
tomerk: added comment.

src/main/scala/internal/AnalysisUtils.scala
tomerk: comment added.

shivaram:
@tomerk really nice work with the design doc. I just left a few comments inline for things that could use some clarification. I haven't taken a detailed look at the logic of each of the methods used, but I like the API design and if we have test cases that go through all the code paths, I'd be fine with merging this.

etrain:
I ran scoverage (https://github.com/scoverage/sbt-scoverage) against this PR and it looks like the unit tests basically caught everything but
1. The `Operator.label` method isn't really called anywhere.
2. `Graph.nextNodeIds` isn't ever called on an empty graph - is this a case we would run into?
3. There are a bunch of checks in `connectGraph` and `replaceNodes` that we don’t ever feed in invalid input.

I don't particularly care about 1, but we should probably add some checks for 2 (if necessary) and 3.

Otherwise this is awesome!

tomerk:
Okay, just added tests for cases 2 and 3!

PR #271, opened 2016-04-29
Pointing to objects located on S3.

### LINKED ISSUE #268: Put vlfeat on s3
We depend on an oxford website in our build script which went down earlier today. I expect it to come back, but it is worthwhile to create a mirror of it on S3 and depend on that directly.

Thanks for reporting @zhaozhang !

etrain:
Closes #268 

Thanks for reporting @zhaozhang


etrain:
I've verified that I can `wget` both files - merging this.

PR #272, opened 2016-04-29
POS Tagging

ngarneau:
Hey guys, just wrote a little POS Tagger wrapper around the one of Epic's libs and models.
Let me know what you think about this, if it fits your coding style guideline and everything :)

If it's all good I'll go on with the NER.

Cheers,

Nicolas


src/main/scala/nodes/nlp/POSTagger.scala
```diff
@@ -0,0 +1,16 @@
+package nodes.nlp
+
+import epic.sequences.{CRF, Segmentation, TaggedSequence}
+import epic.trees.AnnotatedLabel
+import workflow.Transformer
+import breeze.linalg.Counter2
+import org.apache.spark.rdd.RDD
+
+case class POSTagger(@transient tagger: CRF[AnnotatedLabel, String])
```
etrain: Hmm.. won't this be skipped when the `POSTagger` is serialized as part of a spark job?

etrain: Could you add some doc strings that explain the format of the input and output? E.g. It looks like now it assumes that a "sentence" is input as an `Array[String]` and the return is a "Tagged Sentence" where the part of speech comes first and the original string comes second. Should part-of-speech be an `Enum`? At least in English, we have a limited and well defined set of them.

Also, does it make more sense for the result to have a return type other than `(String, String)`? If I think about this in the context of other NLP methods I wonder if there's a richer data structure that makes sense in that domain.

build.sbt
```diff
@@ -22,9 +22,13 @@ libraryDependencies ++= Seq(
   "org.slf4j" % "slf4j-api" % "1.7.2",
   "org.slf4j" % "slf4j-log4j12" % "1.7.2",
   "org.scalatest" %% "scalatest" % "1.9.1" % "test",
+  "org.mockito" % "mockito-core" % "1.8.5",
```
etrain: I haven't used Mockito personally, but it looks like this is what Spark is using for mock tests these days, so I'm good with adding this as a dependency.

etrain:
Thanks for the contribution, Nicolas!

This looks pretty good and is nice and self-contained. There's one comment in there about serialization concerns - maybe we should just test this to verify that its not an issue?

The other comment about interfaces is a little bit more meta. I worry a little bit about having `String` everywhere in the NLP stack because strings are pretty expensive in the JVM, additionally, we should maybe sketch out how this might be used in a pipeline with other NLP feature extractors. I could see `Sentence` being a class and `AnnotatedSentence` subclassing that. We could support a number of different Annotations, including part-of-speech, named entities, etc.

ngarneau:
Thank you very much for your comments Evan!

I will definitely test with the serialization. I did have some problems with Epic with a unserializable classe so it maybe be a concern and worth checking it.

I don't have any problem using the `AnnotatedSentence` class, I'll fix that too.

Thanks again for your comments it's exactly the kind of hints I was looking for and be fixing everything up real soon :)

ngarneau:
I tested the POS model on Spark and it works well. Since it's pretty heavy (~300mb) I also created a case class taking a broadcasted model as parameter. Did not test it on a Spark cluster but it may help with several executors.

I also used the TaggedSequence offered by the Epic lib as a data structure returned by the node. It as richer features to work with.

Let me know if you see anything else :)

Cheers

ngarneau:
Sorry I think I messed up this branch's history, I'm gonna reopen a pull request with clean one.

PR #273, opened 2016-05-01
Pos tagging

ngarneau:
I tested the POS model on Spark and it works well. Since it's pretty heavy (~300mb) I also created a case class taking a broadcasted model as parameter. Did not test it on a Spark cluster but it may help with several executors.

I also used the TaggedSequence offered by the Epic lib as a data structure returned by the node. It as richer features to work with.

Let me know if you see anything else :)

Cheers

### DIFF
--- a/src/main/scala/nodes/nlp/POSTagger.scala
+++ b/src/main/scala/nodes/nlp/POSTagger.scala
@@ -28,6 +28,16 @@ case class POSTagger(model: CRF[AnnotatedLabel, String])
   }
 }
 
+/**
+  * Here's an example using it with a broadcast variable:
+  * {{{
+  *   val model = epic.models.PosTagSelector.loadTagger("en").get
+  *   val broadcastModel = sc.broadcast(model)
+  *   val POSTagger = BroadcastPOSTagger(broadcastModel).apply(data)
+  * }}}
+  *
+  * @param model The Broadcasted POS Tagger model loaded from the Epic library
+  */
 case class BroadcastPOSTagger(model: Broadcast[CRF[AnnotatedLabel, String]])
   extends Transformer[Array[String], TaggedSequence[AnnotatedLabel, String]] {
   override def apply(in: Array[String]): TaggedSequence[AnnotatedLabel, String] = {
--- a/src/main/scala/nodes/nlp/POSTagger.scala
+++ b/src/main/scala/nodes/nlp/POSTagger.scala
@@ -21,7 +21,7 @@ import org.apache.spark.rdd.RDD
   *
   * @param model The POS Tagger model loaded from the Epic library
   */
-case class POSTagger(model: CRF[AnnotatedLabel, String])
+case class POSTagger(@transient model: CRF[AnnotatedLabel, String])
   extends Transformer[Array[String], TaggedSequence[AnnotatedLabel, String]] {
 
   def apply(in: Array[String]): TaggedSequence[AnnotatedLabel, String] = {
--- a/src/main/scala/nodes/nlp/POSTagger.scala
+++ b/src/main/scala/nodes/nlp/POSTagger.scala
@@ -3,7 +3,7 @@ package nodes.nlp
 import epic.sequences.{CRF, TaggedSequence}
 import epic.trees.AnnotatedLabel
 import workflow.Transformer
-import org.apache.spark.broadcast.Broadcast
+import org.apache.spark.rdd.RDD
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/scala/nodes/nlp/POSTagger.scala
```diff
@@ -0,0 +1,46 @@
...
+  * {{{
+  *   val model = epic.models.PosTagSelector.loadTagger("en").get
+  *   val POSTagger = POSTagger(model).apply(data)
+  * }}}
+  *
+  * @param model The POS Tagger model loaded from the Epic library
+  */
+case class POSTagger(model: CRF[AnnotatedLabel, String])
+  extends Transformer[Array[String], TaggedSequence[AnnotatedLabel, String]] {
+  override def apply(in: Array[String]): TaggedSequence[AnnotatedLabel, String] = {
```
etrain: I don't see the need for `override` here?

src/main/scala/nodes/nlp/POSTagger.scala
```diff
@@ -0,0 +1,46 @@
...
+  * Here's an example using it with a broadcast variable:
+  * {{{
+  *   val model = epic.models.PosTagSelector.loadTagger("en").get
+  *   val broadcastModel = sc.broadcast(model)
+  *   val POSTagger = BroadcastPOSTagger(broadcastModel).apply(data)
+  * }}}
+  *
+  * @param model The Broadcasted POS Tagger model loaded from the Epic library
+  */
+case class BroadcastPOSTagger(model: Broadcast[CRF[AnnotatedLabel, String]])
```
etrain: If the model is indeed 300mb, then I think broadcasting by default is a good idea. 

To date, we haven't usually had separate transformers for these implementations. Usually what we do is have the `apply(in: T): U` method be broadcast-less. That way a pipeline that is running e.g. as part of a web service call doesn't need to have a handle to a spark cluster and can run everything locally. But, then we'll `override def apply(in: RDD[T]): U` and put the broadcasting logic in there, because there you're clearly working with RDDs and therefore spark. A simple example is here: https://github.com/amplab/keystone/blob/master/src/main/scala/nodes/stats/CosineRandomFeatures.scala

src/main/scala/nodes/nlp/POSTagger.scala
```diff
@@ -0,0 +1,46 @@
...
+  * Here's an example:
+  * {{{
+  *   val model = epic.models.PosTagSelector.loadTagger("en").get
+  *   val POSTagger = POSTagger(model).apply(data)
+  * }}}
+  *
+  * @param model The POS Tagger model loaded from the Epic library
+  */
+case class POSTagger(model: CRF[AnnotatedLabel, String])
+  extends Transformer[Array[String], TaggedSequence[AnnotatedLabel, String]] {
```
etrain: In general I get a little nervous about exposing classes from other-libraries at user-facing interfaces, but this class from the epic API has a reasonably simple definition and has a bunch of the functionality we're after, so I'm ok with this for now. If it becomes a problem in the future, providing a Keystone-specific version of the interface that's API-compatible shouldn't be too hard.

etrain:
Thanks again for this! Other than my comment about broadcasting this LGTM.

src/main/scala/nodes/nlp/POSTagger.scala
ngarneau: Cool, I thought we'd need to broadcast before the call to the `apply` method was made. I'll change that. Thanks for the example.

etrain:
Ah, one more thing - add yourself to `CONTRIBUTORS.md`? :)

src/main/scala/nodes/nlp/POSTagger.scala
```diff
@@ -0,0 +1,36 @@
...
+  *
+  * Here's an example:
+  * {{{
+  *   val model = epic.models.PosTagSelector.loadTagger("en").get
+  *   val POSTagger = POSTagger(model).apply(data)
+  * }}}
+  *
+  * @param model The POS Tagger model loaded from the Epic library
+  */
+case class POSTagger(model: CRF[AnnotatedLabel, String])
```
etrain: Also, now that we're broadcasting this model, I think it makes sense to mark this as `@transient`, because we explicitly handle sending the model to the workers and we don't want the model getting picked up by Spark's closure capture.

ngarneau: Sounds good!

etrain:
LGTM, merging this - and thanks @ngarneau !

PR #274, opened 2016-05-05
Named entity recognition

ngarneau:
Hey guys, on the same idea as the POS tagger, here is the NER. As @etrain said, it would be better if the Keystone API wouldn't expose Epic's interfaces so I'll work on this as a next step.

Let me know if you see anything here, I built it pretty much the same way I did with the POS Tagger.

Thanks,
Nicolas

### DIFF
--- a/src/main/scala/nodes/nlp/NER.scala
+++ b/src/main/scala/nodes/nlp/NER.scala
@@ -1,0 +1,32 @@
+package nodes.nlp
+
+import epic.sequences.{Segmentation, SemiCRF}
+import org.apache.spark.rdd.RDD
+import workflow.Transformer
+
+/**
+  * Transformer that uses Epic's named entity recognition on a sequence of words.
+  * It is recommended to use a Tokenizer before applying the NER,
+  * assuming we get a list of strings as input.
+  *
+  * Here's an example:
+  * {{{
+  *   val model = epic.models.NerSelector.loadNer("en").get
+  *   val NERmodel = NERModel(model).apply(data)
+  * }}}
+  *
+  * @param model The NER model loaded from the Epic library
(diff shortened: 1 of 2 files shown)


src/main/scala/nodes/nlp/NER.scala
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+  * Transformer that uses Epic's named entity recognition on a sequence of words.
+  * It is recommended to use a Tokenizer before applying the NER,
+  * assuming we get a list of strings as input.
+  *
+  * Here's an example:
+  * {{{
+  *   val model = epic.models.NerSelector.loadNer("en").get
+  *   val NERmodel = NERModel(model).apply(data)
```
etrain: this docstring looks broken?

src/main/scala/nodes/nlp/NER.scala
```diff
@@ -0,0 +1,32 @@
...
+  * }}}
+  *
+  * @param model The NER model loaded from the Epic library
+  */
+case class NER(@transient model: SemiCRF[Any, String])
```
etrain: Is `Any` appropriate here? 
why not `case class NER[T](@transient model: SemiCRF[T, String]) extends Transformer[Array[String], Segmentation[T, String]]`?

etrain:
A couple of minor comments - in general we're trying to preserve type-safety at the level of pipeline nodes, and I think a generic will solve that problem here (though it may not - the epic type system isn't as crazy as breeze's but it's not the simplest one either)

ngarneau:
I definitely agree with your comments.

etrain:
LGTM. Thanks!

PR #276, opened 2016-05-25
Internal dag execution

tomerk:
Introduces user APIs that build on top of the previously introduced Graph.


src/main/scala/workflow/graph/Transformer.scala
```diff
@@ -0,0 +1,76 @@
...
+  protected def singleTransform(in: A): B
+
+  /**
+   * The application of this Transformer to an RDD of input items.
+   * This method may optionally be overridden by ML developers.
+   *
+   * @param in The bulk RDD input to pass into this transformer
+   * @return The bulk RDD output for the given input
+   */
+  protected def batchTransform(in: RDD[A]): RDD[B] = in.map(singleTransform)
```
etrain: Why is this not just `apply`?

tomerk: Transformer extends Pipeline, meaning the publicly exposed API method is `apply`. I can add a note about this to the transformer documentation.

tomerk: Or if you're just talking about just the name itself, it would produce a conflict with the method naming in `Pipeline`.

src/main/scala/workflow/graph/EquivalentNodeMergeRule.scala
```diff
@@ -0,0 +1,57 @@
...
+      (plan, executionState)
+    } else {
+      nodeSetsToMerge.filter(_.size > 1).foldLeft((plan, executionState)) {
+        case ((curPlan, curExecutionState), setToMerge) => {
+          // Construct a graph that merges all of the nodes
+          val nodeToKeep = setToMerge.minBy(_.id)
+          val nextGraph = (setToMerge - nodeToKeep).foldLeft(curPlan) {
+            case (partialMergedPlan, nodeToMerge) => {
+              partialMergedPlan
+                .replaceDependency(nodeToMerge, nodeToKeep)
```
etrain: This graph operator chaining stuff is neat-o!

src/main/scala/workflow/graph/Transformer.scala
etrain: What drew my attention to it is that Cacher overrides these methods and so I start to wonder if other Transformers now have to override these new methods?

tomerk: Hmm that's true, I did change the name of the methods (but not the type signature). I had to do this because `apply` now returns a `Pipeline***Out` instead of immediately returning the transformed rdd/single item.

src/main/scala/workflow/graph/PipelineDatasetOut.scala
```diff
@@ -0,0 +1,25 @@
+package workflow.graph
+
+import org.apache.spark.rdd.RDD
+
+/**
+ * This class is a lazy wrapper around the output of a pipeline that was passed an RDD as input.
+ *
+ * Under the hood, it extends [[GraphExecution]] and keeps track of the necessary execution plan.
+ */
+class PipelineDatasetOut[T] private[graph] (executor: GraphExecutor, sink: SinkId, source: Option[(SourceId, RDD[_])])
```
etrain: Hmm.. we need to think of a better name for this thing? E.g. why is the fact that it's output of something so important that it belongs in the name? Also - I'm unfamiliar with this syntax - is it marking the constructor as `private[graph]`?

tomerk: I'm open to suggestions about better names! And yes, the constructor is being marked `private[graph]`.

src/main/scala/workflow/graph/Estimator.scala
```diff
@@ -0,0 +1,63 @@
+package workflow.graph
+
+import org.apache.spark.rdd.RDD
+
+/**
+ * An estimator has a `fitRDD` method which takes an input and emits a [[Transformer]]
```
etrain: So this looks like another change to the programming model? E.g. all Estimators now have to support a fitRDD method? Why can't they just support a `fit` method?

tomerk: This is another case where the type signature is the same, but the method that needs to be implemented was renamed. This change was done so that calling `est.fit(...)` is now lazy, bringing with it all the lazy/incremental pipeline construction stuff we've been discussing for the past few weeks.

src/main/scala/workflow/graph/LabelEstimator.scala
```diff
@@ -0,0 +1,108 @@
+package workflow.graph
+
+import org.apache.spark.rdd.RDD
+
+/**
+ * A LabelEstimator has a `fitRDDs` method which takes input data and input labels,
```
etrain: I'm not sure if this comment belongs here. Does a user who extends `LabelEstimator` need to implement this non-type-safe thing?

etrain: Oh, I see now - you ask them to implement the type-safe thing. I guess this comes back to my naming thing above.

tomerk: The `fitRDDs` in LabelEstimator that needs to be implemented is typesafe:
`protected def fitRDDs(data: RDD[A], labels: RDD[L]): Transformer[A, B]`

What isn't type-safe is the already implemented overridden method from the EstimatorOperator:
```
  final override private[graph] def fitRDDs(inputs: Seq[DatasetExpression]): TransformerOperator = {
    fitRDDs(inputs(0).get.asInstanceOf[RDD[A]], inputs(1).get.asInstanceOf[RDD[L]])
  }
```

src/main/scala/workflow/graph/Identity.scala
```diff
@@ -0,0 +1,15 @@
...
+import org.apache.spark.rdd.RDD
+
+import scala.reflect.ClassTag
+
+/**
+ * This transformer performs a no-op on its input.
+ *
+ * @tparam T Type of the input and, by definition, output.
+ */
+case class Identity[T : ClassTag]() extends Transformer[T,T] {
```
etrain: Similar to cacher - suddenly users have to implement different methods - it's not clear to me why this naming has to change.

src/main/scala/workflow/graph/Estimator.scala
```diff
@@ -0,0 +1,63 @@
...
+    // Because pipeline construction is incremental, we make sure to add the states of the data,
+    // removing the graph ids that no longer exist.
+    val newState = data.getState - data.getSink
+    new ConcretePipeline(new GraphExecutor(newGraph, newState), sourceId, sinkId)
+  }
+
+  /**
+   * The non-type-safe `fitRDDs` method of [[EstimatorOperator]] that is being overridden by the Estimator API.
+   */
+  final override private[graph] def fitRDDs(inputs: Seq[DatasetExpression]): TransformerOperator = {
```
etrain: I'm not following why this is needed - but I guess it's because I'm not following the dataflow here..

src/main/scala/workflow/graph/GatherTransformer.scala
```diff
@@ -0,0 +1,19 @@
+package workflow.graph
+
+import org.apache.spark.rdd.RDD
+
+/**
+ * A [[TransformerOperator]] that gathers multiple datasets of {@tparam T} into a dataset of Seq[T]
+ * (Or individual datums of T into a single Seq[T])
+ */
+private[workflow] case class GatherTransformer[T]() extends TransformerOperator {
+  override private[graph] def singleTransform(inputs: Seq[DatumExpression]): Any = {
```
etrain: Should the output of this be typed as `Any` or as `T`?

src/main/scala/workflow/graph/GatherTransformer.scala
```diff
@@ -0,0 +1,19 @@
...
+    inputs.map(_.get.asInstanceOf[T])
+  }
+
+  override private[graph] def batchTransform(inputs: Seq[DatasetExpression]): RDD[_] = {
```
etrain: Same as above.

src/main/scala/workflow/graph/GatherTransformer.scala
tomerk: This is an untyped method we're overriding, so `Any` is correct.

src/main/scala/workflow/graph/GatherTransformer.scala
tomerk: Same as above. The methods in TransformerOperator we're overriding are untyped, so returning `RDD[_]` is fine, but if we wanted to we could type it as `RDD[Seq[T]]` because  `RDD[Seq[T]] <: RDD[_]`

src/main/scala/workflow/graph/GraphExecution.scala
```diff
@@ -0,0 +1,66 @@
+package workflow.graph
+
+// rough idea given incrementalism: do everything it can in the base executor (which may be shared w/ other things) w/o inserting sources.
+// then create a "final executor" that is the base one w/ sources inserted, and optimized using the EquivalentNodeMerge optimizer.
+// The final value execution happens on that "final executor"
+// This two stage process allows "intuitive things" to happen a source being passed in is already processed elsewhere in the pipeline (e.g. making sure to reuse a cacher),
+// while pipeline fitting results can be reused across multiple pipeline applies, as they all share the same base executor.
+abstract class GraphExecution[T](
```
etrain: More docs here would be cool - I really don't fully grok what this thing is

src/main/scala/workflow/graph/GraphExecution.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  private[graph] def getSources: Map[SourceId, Operator] = _sources
+  private[graph] def getSink: SinkId = _sink
+
+  private var ranExecution: Boolean = false
+  private lazy val finalExecutor: GraphExecutor = {
+    if (getSources.nonEmpty) {
+      _executor.partialExecute(getSink)
+
+      val (newGraph, newState) = EquivalentNodeMergeOptimizer.execute(getGraph, getState)
```
etrain: Hmm.. I'm not following but why is this thing being run here?

src/main/scala/workflow/graph/GraphExecution.scala
```diff
@@ -0,0 +1,66 @@
...
+      new GraphExecutor(newGraph, newState, optimize = false)
+    } else {
+      _executor
+    }
+  }
+
+  private[graph] def getGraph: Graph = if (ranExecution) {
+    finalExecutor.getGraph
+  } else {
+    _sources.foldLeft(_executor.getGraph) {
```
etrain: Would be really helpful to explain this execution logic in words.

src/main/scala/workflow/graph/GraphExecution.scala
tomerk: Yeah, this is the last class I'm still working on docs for.

src/main/scala/workflow/graph/GraphExecutor.scala
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * Warning: Not thread-safe!
+ *
+ * @param graph The underlying graph to execute
+ * @param state A mapping of GraphId to computed Expression.
+ *              This is the current state of what has already been executed in the graph.
+ * @param optimize Whether or not to optimize the graph using the global executor before executing anything.
+ *                 Defaults to True.
+ */
+private[graph] class GraphExecutor(graph: Graph, state: Map[GraphId, Expression], optimize: Boolean = true) {
```
etrain: Names are really confusing - what's the difference between a `GraphExecution` and a `GraphExecutor`?

src/main/scala/workflow/graph/GraphExecution.scala
tomerk: So here is an explanation at a high level while I write the docs for this class.
In order for pipeline fitting to be carried through across `pipeline.apply` calls, and to ensure common subexpressions get eliminated when we do something like pass training data into a pipeline (when there is a cacher in the featurization steps), we do this GraphExecution as a two step process:

The first fits/optimizes/partially executes the GraphExecutor shared across all `pipeline.apply` calls as far as possible BEFORE inserting any sources. Then, it inserts the sources into the graph, does common subexpression elimination, and produces a `finalExecutor` used ONLY by that specific `pipeline.apply(data)` execution.

This two-step process allows us to fit a pipeline and share the fitted results across all `pipeline.apply` calls, AND not do unnecessary work when a pipeline is applied to it's own training data/ something we would reasonably expect to have been cached.

etrain: Specifically - why are we running the `EquivalentNodeMergeOptimizer` here. It seems out of place and perhaps like a leaky abstraction?

src/main/scala/workflow/graph/GraphExecution.scala
tomerk: Sure, working on the documentation for this class still!

src/main/scala/workflow/graph/GraphExecutor.scala
tomerk: GraphExecutor: Given a graph & execution state, provides methods that will optimize the graph & execute it up to arbitrary GraphIds.

GraphExecution: Lazily represent the result of executing a pipeline on data. Given a GraphExecutor, a desired sink, and optionally sources that still need to be inserted into the graph, do the two-step process I describe above to get the result of executing the pipeline on data.

tomerk: New name suggestions are welcome!

etrain:
I am still working my way through this PR and won't look at your responses tonight - so don't feel like you need to post them real-time. 

My meta-takeaway from this so far is that 
1. There are a bunch of new classes being introduced, and I don't have a clear sense of the execution flow through these new classes - I don't know if it's a written description or a box and pointers diagram or what - but as it stands now this code feels like something that fits in your head but isn't self-explanatory. It doesn't have to be self-explanatory, of course, but it should be something you can explain in a small document or ideally on a napkin :)
2. Let's be careful about changing our user-facing APIs too much. They're simple and that's a good thing. Changing method names that we expose to users when simply creating a new method name for the internal behavior we want to support is probably a better approach.
3. I think there's a decent amount of confusion/complexity coming in because we say that pipelines take RDDs and emit `PipelineDatasetOut`. Instead if we're going to introduce yet another wrapper for a data set, we should at least have the API consistently work with these things both as input and output. I'm still not convinced we need it, but it will at least make the API more consistent and easier to understand and may solve a bunch of the naming/override problems we're seeing.

src/main/scala/workflow/graph/PipelineDatumOut.scala
```diff
@@ -0,0 +1,23 @@
+package workflow.graph
+
+/**
+ * This class is a lazy wrapper around the output of a pipeline that was passed a single datum as input.
+ *
+ * Under the hood, it extends [[GraphExecution]] and keeps track of the necessary execution plan.
+ */
+class PipelineDatumOut[T] private[graph] (executor: GraphExecutor, sink: SinkId, source: Option[(SourceId, Any)])
```
etrain: Also needs a better name :)

src/main/scala/workflow/graph/Transformer.scala
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  final override private[graph] def batchTransform(inputs: Seq[DatasetExpression]): RDD[_] = {
+    batchTransform(inputs.head.get.asInstanceOf[RDD[A]])
+  }
+}
+
+object Transformer {
+  /**
+   * This constructor takes a function and returns a Transformer that maps it over the input RDD
```
etrain: Looks like this thing is meant to support inlining of transformers that are regular scala functions. You should say as much in the docstring.

tomerk: I just copied & pasted the existing documentation for this method. I'll add the note!

PR #277, opened 2016-05-26
Prefix-based execution

tomerk:


etrain:
For the purposes of documentation - this PR is the result of some extended discussions between myself, @tomerk and @shivaram regarding expected behavior for a pipeline developer when interacting with the system. Here we are introducing the concept of a "logical prefix" and a table of logical prefixes which effectively serves as a lookup table for work that has already been done on a pipeline in a way that is robust to pipeline optimization. It is designed to supplant #276.

src/main/scala/workflow/graph/Cacher.scala
```diff
@@ -0,0 +1,24 @@
...
+import scala.reflect.ClassTag
+
+/**
+ * Caches an RDD at a given point within a Pipeline. Follows Spark's lazy evaluation conventions.
+ *
+ * @param name An optional name to set on the cached output. Useful for debugging.
+ * @tparam T Type of the input to cache.
+ */
+case class Cacher[T: ClassTag](name: Option[String] = None) extends Transformer[T,T] with Logging {
+  override protected def batchTransform(in: RDD[T]): RDD[T] = {
```
etrain: So that we don't forget my comments on #276 - we really need to get the names right for Transformer and Estimator here.

src/main/scala/workflow/graph/DanglingNodeRemovalRule.scala
```diff
@@ -0,0 +1,23 @@
+package workflow.graph
+
+/**
+ * A rule to remove all nodes & sources in a graph that don't lead to any sink.
+ */
+object DanglingNodeRemovalRule extends Rule {
```
etrain: How about `UnusedNodeRemovalRule`.

etrain: Or better `UnusedBranchRemovalRule`?

src/main/scala/workflow/graph/Prefix.scala
```diff
@@ -0,0 +1,25 @@
...
+      case dep: NodeId => get(graph, dep)
+      case dep: SourceId =>
+        throw new IllegalArgumentException("May not get the prefix of a node with Sources in the dependencies.")
+    }
+
+    Prefix(rootOp, deps)
+  }
+}
+
+private[graph] case class Prefix(operator: Operator, deps: Seq[Prefix])
```
etrain: Can we add some docs here? It looks to me like a prefix is a tree of operators, and the leaves of the tree just have empty `deps` - right?

Also - a case class normally has a Companion object associated with it - but here you have already defined a companion object - what happens in this case? E.g. do we not have a `Prefix.apply()` or something?

src/main/scala/workflow/graph/ExtractSaveablePrefixes.scala
```diff
@@ -0,0 +1,19 @@
+package workflow.graph
+
+/**
+ * Extract the prefixes whose state we want to save (aka for any cacher or estimator node)
```
etrain: This basically looks ok. It kind of seems like this set of rules is somehow special?

etrain: That is, with respect to assumptions they make about their inputs.

src/main/scala/workflow/graph/Rule.scala
```diff
@@ -0,0 +1,19 @@
...
+ * attached to the old DAG.
+ */
+abstract class Rule {
+  /** Name for this rule, automatically inferred based on class name. */
+  val ruleName: String = {
+    val className = getClass.getName
+    if (className endsWith "$") className.dropRight(1) else className
+  }
+
+  def apply(plan: Graph, prefixes: Map[NodeId, Prefix]): (Graph, Map[NodeId, Prefix])
```
etrain: It's a little weird that *all* rules need this `prefixes` map.

src/main/scala/workflow/graph/Pipeline.scala
```diff
@@ -0,0 +1,278 @@
...
+    labels: PipelineDatasetOut[L]
+  ): Pipeline[A, C] = {
+    this andThen est.fit(apply(data), labels)
+  }
+
+}
+
+object Pipeline {
+  // not threadsafe
+  private[graph] val state: scala.collection.mutable.Map[Prefix, Expression] = scala.collection.mutable.Map()
```
etrain: Need more comments on what this thing is.

src/main/scala/workflow/graph/SavedStateLoadRule.scala
```diff
@@ -0,0 +1,19 @@
+package workflow.graph
+
+/**
+ * A rule to load any saved state for the nodes w/ prefixes attached
+ */
+object SavedStateLoadRule extends Rule {
+  override def apply(plan: Graph, prefixes: Map[NodeId, Prefix]): (Graph, Map[NodeId, Prefix]) = {
+    val newGraph = prefixes.foldLeft(plan) {
```
etrain: Please document what this thing is actually doing. 

My understanding is this - for each "interesting" prefix that exists in the pipeline (we know that the input to this function is only a set of prefixes that exist in the pipeline), check to see whether it is in the state map. If it is, replace the prefix with an ExpressionOperator which is either a fitted transformer or a cached rdd - correct?

src/main/scala/workflow/graph/DanglingNodeRemovalRule.scala
```diff
@@ -0,0 +1,23 @@
...
+  override def apply(plan: Graph, prefixes: Map[NodeId, Prefix]): (Graph, Map[NodeId, Prefix]) = {
+    val ancestorsOfSinks = plan.sinks.foldLeft(Set[GraphId]()) {
+      case (ancestors, sink) => ancestors ++ AnalysisUtils.getAncestors(plan, sink)
+    }
+
+    val nodesToRemove = plan.nodes -- ancestorsOfSinks.collect { case node: NodeId => node }
+    val sourcesToRemove = plan.sources -- ancestorsOfSinks.collect { case source: SourceId => source }
+
+    val afterSourceRemoval = sourcesToRemove.foldLeft(plan) {
+      case (curPlan, sourceToRemove) => curPlan.removeSource(sourceToRemove)
```
etrain: again, these graph operators are awesome!

src/main/scala/workflow/graph/DanglingNodeRemovalRule.scala
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    nodesToRemove.foldLeft((afterSourceRemoval, prefixes)) {
+      case ((curPlan, curPrefixes), nodeToRemove) => (curPlan.removeNode(nodeToRemove), curPrefixes - nodeToRemove)
```
etrain: I'm a little worried that maintaining this prefix state is going to come back to bit us, but I'm willing to live with it.

src/main/scala/workflow/graph/EquivalentNodeMergeRule.scala
```diff
@@ -0,0 +1,47 @@
...
+ * Nodes are considered equivalent if:
+ * - The operators stored within the nodes are equal (.equals() is true)
+ * - They share the same dependencies
+ *
+ * This rule also merges prefixes if any of
+ * the nodes being merged have their prefix attached.
+ */
+object EquivalentNodeMergeRule extends Rule {
+  override def apply(plan: Graph, prefixes: Map[NodeId, Prefix]): (Graph, Map[NodeId, Prefix]) = {
+    val nodeSetsToMerge = plan.nodes.groupBy(id => (plan.getOperator(id), plan.getDependencies(id))).values
```
etrain: I guess one obvious question is why this rule doesn't operate in terms of prefixes. But, given that it works in this form I don't see a real reason to change it now.

src/main/scala/workflow/graph/EquivalentNodeMergeRule.scala
```diff
@@ -0,0 +1,47 @@
...
+          val nodeToKeep = setToMerge.minBy(_.id)
+          val nextGraph = (setToMerge - nodeToKeep).foldLeft(curPlan) {
+            case (partialMergedPlan, nodeToMerge) => {
+              partialMergedPlan
+                .replaceDependency(nodeToMerge, nodeToKeep)
+                .removeNode(nodeToMerge)
+            }
+          }
+
+          // If any of the nodes being merged have been executed, update the prefixes
```
etrain: I am having a bit of a tough time following why this node should need to know about prefixes, or why prefixes should be something considered after the first stage of optimization at all? 

If we combine the first two rules into a single rule, and eliminate "prefixes" from the `Rule` interface, do we get to kill this code?

Are we giving anything up other than chaining other rules that might need prefixes (which don't yet exist)? Or is there something I'm not considering with respect to carrying results of optimization forward or something?

src/main/scala/workflow/graph/DefaultOptimizer.scala
```diff
@@ -0,0 +1,11 @@
+package workflow.graph
+
+/**
+ * The default Pipeline optimizer used when executing pipelines.
+ */
+object DefaultOptimizer extends Optimizer {
+  protected val batches: Seq[Batch] =
+    Batch("Load Saved State", Once, ExtractSaveablePrefixes, SavedStateLoadRule, DanglingNodeRemovalRule) ::
```
etrain: If we just combine ExtractSavablePrefixes and SavedStateLoadRule into a single rule do we eliminate the need for prefixes in other rules?

src/main/scala/workflow/graph/Estimator.scala
```diff
@@ -0,0 +1,60 @@
...
+    fitRDD(inputs.head.get.asInstanceOf[RDD[A]])
+  }
+
+  /**
+   * The type-safe method that ML developers need to implement when writing new Estimators.
+   *
+   * @param data The estimator's training data.
+   * @return A new transformer
+   */
+  protected def fitRDD(data: RDD[A]): Transformer[A, B]
```
etrain: So again I'd rather have this be called `fit` (since this is what the developers are already implementing) and the other methods called something else. E.g. small change means no change to existing estimators. Does that cause some naming collisions that I'm not anticipating?

src/main/scala/workflow/graph/GatherTransformer.scala
```diff
@@ -0,0 +1,19 @@
+package workflow.graph
+
+import org.apache.spark.rdd.RDD
+
+/**
+ * A [[TransformerOperator]] that gathers multiple datasets of {@tparam T} into a dataset of Seq[T]
+ * (Or individual datums of T into a single Seq[T])
+ */
+private[workflow] case class GatherTransformer[T]() extends TransformerOperator {
```
etrain: I was confused that this was called `GatherTransformer` mind switching it to `GatherTransformerOperator` - more verbose but its not exactly a regular transformer.

src/main/scala/workflow/graph/DefaultOptimizer.scala
tomerk: We need the correct prefix mapping available post-optimization, at execution time. This is to save the result of a node in the global state w/ the correct prefix. This is why every rule needs to maintain the prefix mappings for the nodes whose results we will want to save.

etrain: This seems fishy to me - but I'll come back to this after I review the execution logic.

src/main/scala/workflow/graph/Estimator.scala
tomerk: Yes, this is still the same situation as all the rest. Because we do extra lazy wrapper magic, this returns a different output type than the 'fit' method pipeline chainers use. Meaning, this method needed to be renamed or else there's two methods w/ the same name and input types, and it won't compile.

src/main/scala/workflow/graph/GraphExecutor.scala
```diff
@@ -0,0 +1,81 @@
...
+ * @param graph The underlying graph to execute
+ * @param optimize Whether or not to optimize the graph using the global executor before executing anything.
+ *                 Defaults to True.
+ */
+private[graph] class GraphExecutor(val graph: Graph, optimize: Boolean = true) {
+  // internally used flag to check if optimization has occurred yet. Not thread-safe!
+  private var optimized: Boolean = false
+
+  // The lazily computed result of optimizing the initial graph w/ initial state attached to it
+  private lazy val optimizedGraphAndPrefixes: (Graph, Map[NodeId, Prefix]) = {
```
etrain: Ah, ok - I get it. We need this map of NodeId -> Prefix so that when we actually execute a section of the graph (that we might care about later) we make sure that the result is in the state table?

I'll try to think of another way to do this, but I guess this works.

src/main/scala/workflow/graph/Operator.scala
```diff
@@ -162,3 +162,13 @@ private[graph] class DelegatingOperator extends Operator with Serializable {
...
 }
+
+/**
+ * An operator initialized with an [[Expression]], which takes no inputs and always outputs
+ * that expression, regardless of the inputs.
+ *
+ * @param expression The expression to always output
+ */
+private[graph] class ExpressionOperator(expression: Expression) extends Operator {
+  override def execute(deps: Seq[Expression]): Expression = expression
```
etrain: I wonder if it makes sense to `assert` that deps is empty here?

src/main/scala/workflow/graph/Estimator.scala
etrain: Would not having to deal with RDD anywhere solve some of these naming issues? E.g. we require that a user hand us a PipelineDataset (which has a `.get` that yields the RDD and some convenience constructor that just creates a thing that holds an RDD)?

etrain: Thinking that through a little bit I'm guessing that would be pretty big changes for this PR. I guess I'm back to wondering why the existing user interfaces change instead of these new ones getting new names? Maybe this is a whiteboard discussion

tomerk: Unfortunately that's not enough, because if we let the user create a PipelineDataset it wouldn't necessarily hold the whole pipeline DAG if they did it wrong (and making having lazy results be totally pointless)

src/main/scala/workflow/graph/PipelineResult.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  private[graph] def setExecutor(executor: GraphExecutor): Unit = {
+    this._executor = executor
+  }
+
+  private[graph] def setSink(sink: SinkId): Unit = {
+    this._sink = sink
+  }
+
+  private[graph] def getSink: SinkId = _sink
```
etrain: Declaring these as `private[graph] var` in the constructor doesn't get you getters and setters?

src/main/scala/workflow/graph/PipelineResult.scala
```diff
@@ -0,0 +1,30 @@
+package workflow.graph
+
+// rough idea given incrementalism: do everything it can in the base executor (which may be shared w/ other things) w/o inserting sources.
```
etrain: Again, good documentation here would be highly appreciated.

src/main/scala/workflow/graph/Pipeline.scala
```diff
@@ -0,0 +1,278 @@
...
+   * their underlying execution plans together, instead of optimizing each plan individually.
+   * When a workload consists of processing multiple datasets using multiple pipelines,
+   * this may produce superior workload throughput.
+   *
+   * This is a lazy method, i.e. no optimization or execution will occur until the submitted executions have
+   * their results accessed.
+   *
+   * @param executions The executions to submit as a group.
+   */
+  def submit(executions: PipelineResult[_]*): Unit = {
```
etrain: Remind me to think about whether or not we should keep this.

src/main/scala/workflow/graph/Estimator.scala
etrain: I mean that a user creates a `PipelineDataset` which is just and rdd - and
the graph would be empty other than the RDD?

etrain:
Alright - I've taken a pretty close look at this, and I think it's cleaner/easier to reason about than #276. I am a little worried that carrying prefixes around the way we are is confusing. Also, I'm not completely happy with the PipelineResult/PipelineDatasetOut, etc. business. It somehow feels like more than one layer of indirection at present.

I am also a little concerned that we are adding additional complexity to `Transformer` and `Estimator` and that maybe the relationship between `TransformerOperator` and `Transformer` should flip - e.g. make Transformer a member of TransformerOperator - that way `TransformerOperator` *has* a Transformer rather than `Transformer` *is a* `TransformerOperator`. I realize that this bit of things comes from our current master.

For now, I guess I would say try and figure out how to resolve the naming situation. 

Allow me to summarize what users can expect from this interface:

1. Pipelines are constructed using the usual `andThen` syntax.
2. When a pipeline is `apply`'ed to data, the pipeline returns a lazy object (`PipelineResult`) which can be passed to new pipelines as input, or explicitly "materialized" (air quotes because these things are often lazy RDDs). 
3. When a user calls `get` to materialize the pipeline output, the pipeline is optimized and executed. Up to that point - the result of any `Cacher` or `Estimator` nodes are put into a global table for later lookup.
4. If a user runs a pipeline with the same logical prefix (that generated the result of a Cacher or Estimator) anywhere in it, the global table will be consulted and the pipeline will be modified to use this cached state. Optimization will be done without knowledge of the prefix (that is, we're assuming that reading cacher results is at least as fast as re-running the pipeline.

I'm not 100% convinced that this is a "fits in the users head" kind of thing - but at least the current iteration of this PR (mostly) fits in my head 👍

src/main/scala/workflow/graph/Chainable.scala
```diff
@@ -0,0 +1,119 @@
+package workflow.graph
+
+import org.apache.spark.rdd.RDD
+
+/**
+ * TODO Write me :(
```
etrain: Please to this :)

src/main/scala/workflow/graph/FittedPipeline.scala
```diff
@@ -0,0 +1,33 @@
+package workflow.graph
+
+import org.apache.spark.rdd.RDD
+
+
+class FittedPipeline[A, B] private[graph] (
```
etrain: Some very explicit ScalaDoc here as well.

src/main/scala/workflow/graph/Identity.scala
```diff
@@ -10,6 +10,6 @@ import scala.reflect.ClassTag
  * @tparam T Type of the input and, by definition, output.
  */
 case class Identity[T : ClassTag]() extends Transformer[T,T] {
-  override protected def singleTransform(in: T): T = in
-  override protected def batchTransform(in: RDD[T]): RDD[T] = in
+  override protected def apply(in: T): T = in
```
etrain: You have no idea how much happier I am now :)

src/main/scala/workflow/graph/Transformer.scala
```diff
@@ -15,26 +15,17 @@ import scala.reflect.ClassTag
...
-   * The source this transformer contains under the hood (because it extends Pipeline)
-   */
-  override private[graph] val source = SourceId(0)
-
-  /**
-   * The sink this transformer contains under the hood (because it extends Pipeline)
-   */
-  override private[graph] val sink = SinkId(0)
+abstract class Transformer[A, B : ClassTag] extends TransformerOperator with Chainable[A, B] {
+  private[graph] override def toPipeline: Pipeline[A, B] = new Pipeline(
```
etrain: So if I implement `Chainable` I have to know how to convert myself to a pipeline - is that the right interpretation? And that's it?

tomerk: Yes, that is correct

src/main/scala/workflow/graph/TransformerGraph.scala
```diff
@@ -0,0 +1,18 @@
+package workflow.graph
+
+// This is serializable :)
```
etrain: moar docstring

src/main/scala/workflow/graph/Pipeline.scala
```diff
@@ -13,13 +13,48 @@ import scala.reflect.ClassTag
...
+
+    val newGraph = UnusedBranchRemovalRule.apply(graphWithFitEstimators, Map())._1
+
+    val transformerGraph = TransformerGraph(
+      newGraph.sources,
+      newGraph.sinkDependencies,
+      newGraph.operators.map(op => (op._1, op._2.asInstanceOf[TransformerOperator])),
+      newGraph.dependencies)
+
+    new FittedPipeline[A, B](transformerGraph, source, sink)
```
etrain: Can you give a brief explanation for why `TranformerGraph` exists? What's wrong with constructing it as a regular `Graph`?

tomerk: `Operator` isn't guaranteed to be serializable, but `TransformerOperator` is. I'll note that in the documentation.

src/main/scala/workflow/graph/PipelineDatasetOut.scala
```diff
@@ -0,0 +1,23 @@
...
+ * This class is a lazy wrapper around the output of a pipeline that was passed an RDD as input.
+ *
+ * Under the hood, it extends [[PipelineResult]] and keeps track of the necessary execution plan.
+ */
+class PipelineDatasetOut[T] private[graph] (executor: GraphExecutor, sink: SinkId)
+  extends PipelineResult[RDD[T]](
+    executor,
+    sink)
+
+object PipelineDatasetOut {
```
etrain: I'm thinking we should just call this `PipelineDataset`.

src/main/scala/workflow/graph/PipelineDatumOut.scala
```diff
@@ -0,0 +1,21 @@
+package workflow.graph
+
+/**
+ * This class is a lazy wrapper around the output of a pipeline that was passed a single datum as input.
+ *
+ * Under the hood, it extends [[PipelineResult]] and keeps track of the necessary execution plan.
+ */
+class PipelineDatumOut[T] private[graph](executor: GraphExecutor, sink: SinkId)
```
etrain: ditto here for `PipelineDatum`

src/main/scala/workflow/graph/Cacher.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import scala.reflect.ClassTag
+
+/**
+ * Caches an RDD at a given point within a Pipeline. Follows Spark's lazy evaluation conventions.
+ *
+ * @param name An optional name to set on the cached output. Useful for debugging.
+ * @tparam T Type of the input to cache.
+ */
+case class Cacher[T: ClassTag](name: Option[String] = None) extends Transformer[T,T] with Logging {
```
shivaram: quick question: does this mean we should remove the old cacher ? the code looks identical though to the current one ?

tomerk: It uses workflow.graph.Transformer instead of workflow.Transformer.
When we do the swap, we will need to remove the old one.





On Tue, May 31, 2016 at 6:06 PM -0700, "Shivaram Venkataraman" <notifications@github.com> wrote:












In src/main/scala/workflow/graph/Cacher.scala:
> @@ -0,0 +1,24 @@
> +package workflow.graph
> +
> +import org.apache.spark.rdd.RDD
> +import pipelines.Logging
> +
> +import scala.reflect.ClassTag
> +
> +/**
> + * Caches an RDD at a given point within a Pipeline. Follows Spark's lazy evaluation conventions.
> + *
> + * @param name An optional name to set on the cached output. Useful for debugging.
> + * @tparam T Type of the input to cache.
> + */
> +case class Cacher[T: ClassTag](name: Option[String] = None) extends Transformer[T,T] with Logging {



quick question: does this mean we should remove the old cacher ? the code looks identical though to the current one ?

src/main/scala/workflow/graph/EquivalentNodeMergeRule.scala
shivaram: `groupBy` builds a hash map in scala if I'm not wrong and hash maps are grouped by calling `.hashCode` and not `.equals` -- So we should either change the doc string above to say we use `.hashCode` (does every operator have one ?) or change the code to explicitly use `.equals` ?

src/main/scala/workflow/graph/EquivalentNodeMergeRule.scala
shivaram: Was this resolved offline ?

src/main/scala/workflow/graph/Estimator.scala
```diff
@@ -0,0 +1,60 @@
...
+   */
+  final def withData(data: PipelineDatasetOut[A]): Pipeline[A, B] = {
+    // Remove the data sink,
+    // Then insert this estimator into the graph with the data as the input
+    val curSink = data.executor.graph.getSinkDependency(data.sink)
+    val (estGraph, estId) = data.executor.graph.removeSink(data.sink).addNode(this, Seq(curSink))
+
+    // Now that the estimator is attached to the data, we need to build a pipeline DAG
+    // that applies the fit output of the estimator. We do this by creating a new Source in the DAG,
+    // Adding a delegating transformer that depends on the source and the label estimator,
```
shivaram: Nit: Can we move the comment next to the appropriate line of code in the next 4 lines ?

src/main/scala/workflow/graph/EquivalentNodeMergeRule.scala
tomerk: That's true, scala hashmaps do use .hashCode as well as .equals. I'll update the documentation. So far all operators are either case classes or `object`s, so no issue there yet.

src/main/scala/workflow/graph/ExtractSaveablePrefixes.scala
```diff
@@ -0,0 +1,20 @@
+package workflow.graph
+
+/**
+ * Extract the prefixes of all Nodes whose state we want to save for reuse by other Pipeline apply and fit calls.
+ * This is all nodes that either have a Cacher or an EstimatorOperator as the internal operator.
+ */
+object ExtractSaveablePrefixes extends Rule {
```
shivaram: Just to make sure, for the `Cacher` even if Spark evicts the RDDs this just means that we will still operate with the RDDs on which cache was called. Is that right ?

tomerk: Yup, that's right!

src/main/scala/workflow/graph/GatherTransformer.scala
shivaram: +1

src/main/scala/workflow/graph/GraphExecutor.scala
```diff
@@ -0,0 +1,81 @@
...
+          val dependencies = optimizedGraph.getDependencies(node)
+          val depExpressions = dependencies.map(dep => execute(dep))
+          val operator = optimizedGraph.getOperator(node)
+          val expression = operator.execute(depExpressions)
+
+          // Save state if the prefix is being considered for saving/loading state!
+          // Note: This whole process isn't threadsafe & things could get messed up if
+          // multiple pipelines are executing and/or optimizing at once
+          if (prefixes.contains(node)) {
+            Pipeline.state(prefixes(node)) = expression
```
shivaram: yikes - we have a global variable ? hmm can't we have this information in a map either in this class or in some other class that knows about various pipelines ?

src/main/scala/workflow/graph/Pipeline.scala
```diff
@@ -0,0 +1,169 @@
...
+ * @tparam B type of the data this Pipeline outputs
+ */
+class Pipeline[A, B] private[graph] (
+  private[graph] val executor: GraphExecutor,
+  private[graph] val source: SourceId,
+  private[graph] val sink: SinkId) extends Chainable[A, B] {
+
+  private[graph] def toPipeline: Pipeline[A, B] = this
+
+  // TODO: Add docs
```
shivaram: ?

src/main/scala/workflow/graph/GraphExecutor.scala
tomerk: Ah this is the global prefix state solution @etrain and I ended up going with. If you prefer we can isolate the global state variable to only be exposed by methods in the `Pipeline` object instead of being available as a public val.

tomerk: Or we could encapsulate it so `Pipeline objects` takes a mutable state as a parameter that is shared by multiple pipelines (and somehow come up w/ a good solution for the constructors)

src/main/scala/workflow/graph/PipelineDatasetOut.scala
shivaram: If this is going to be user / developer visible I like `PipelineDataset` better. If not it matters less

src/main/scala/workflow/graph/Prefix.scala
```diff
@@ -0,0 +1,30 @@
...
+   */
+  def get(graph: Graph, node: NodeId): Prefix = {
+    val rootOp = graph.getOperator(node)
+    val deps = graph.getDependencies(node).map {
+      case dep: NodeId => get(graph, dep)
+      case dep: SourceId =>
+        throw new IllegalArgumentException("May not get the prefix of a node with Sources in the dependencies.")
+    }
+
+    Prefix(rootOp, deps)
```
shivaram: if this is the only way to construct a prefix i'd suggest making this a `apply` and making the case class below a normal class.

src/main/scala/workflow/graph/Rule.scala
shivaram: was this resolved ?

src/main/scala/workflow/graph/Prefix.scala
tomerk: case class provides us with `hashCode` and `equals` methods (which end up being used). If you prefer I could manually implement those and make it not a case class.

shivaram:
I took one pass over the PR and this is reasonable clean and I think I can wrap my head around most of it. It might be interesting to write a post-hoc design doc that clarifies where we ended up on the design and why we did so.

Most of my comments are minor - the only major one is the one about `Pipeline.state`

src/main/scala/workflow/graph/Pipeline.scala
tomerk: whoops, forgot about this method's docs!

src/main/scala/workflow/graph/Prefix.scala
shivaram: No strong preference here - its just weird to construct a `Prefix` object using `.get` instead of a `.apply`

src/main/scala/workflow/graph/GraphExecutor.scala
shivaram: So we want multiple pipeline objects to share state - thats the problem statement right ? Do these pipelines share any other objects ? i.e. say the GraphExecutor or some other central object ? Or to put it another way are these pipelines say constructed from one another ? Then we could pass around some shared PipelineState variable in teh constructor ?

src/main/scala/workflow/graph/Prefix.scala
tomerk: Hmm, my thinking was it's not exactly 'constructing' the `Prefix` data structure object, it's 'searching the graph' for the `Prefix` object. If you prefer `apply` though, I can make the method `apply` instead of `get` without having to make `Prefix` not a case class. (As the type signature won't collide)

shivaram: If its searching then I'd just call this `findPrefix` and that should be fine.

src/main/scala/workflow/graph/GraphExecutor.scala
tomerk: Yes, the problem statement is that we want multiple pipeline objects to share state. They do not currently share any other object. The tricky part about trying to pass around shared `PipelineState` via the constructors in the chaining methods is that a Pipeline can be produced by chaining two other Pipelines, and the state between all three pipelines has to become mutably linked (as well as feeding through to all pipelines that were chained to produce those pipelines).

src/main/scala/workflow/graph/Prefix.scala
tomerk: Done.

src/main/scala/workflow/graph/GraphExecutor.scala
shivaram: Hmm I see. It would be easy to create this shared datastructure if we had an entry point for keystone - i.e something like SparkContext. 

In the absence of that how about having something like 
- A global PipelineEnv object which supports a getOrCreate method
- A PipelineEnv class that contains a `pipelineState` object
- This object can be the hashMap itself or a wrapper

To access global state you do something like `PipelineEnv.getOrCreate.pipelineState(blah)`

The main advantage is that we have scoped our global variable to one object now and its not a loose collection of things. Ideally we should also put the optimizer inside this. Any thoughts on this ?

tomerk: That sounds reasonable. Should I also make Pipelines and optimizer rules take a PipelineEnv as a parameter? That way they don't have to call the global `PipelineEnv.getOrCreate`. Or would you prefer that in a later PR?

shivaram: If we can pass things through a constructor thats cool. If thats going to be a bigger change I'm fine with the `getOrCreate` for this PR and then do a follow up PR

tomerk: Okay, I'll push it off for a follow up PR.

src/main/scala/workflow/graph/EquivalentNodeMergeRule.scala
tomerk: updated docs.

src/main/scala/workflow/graph/Chainable.scala
```diff
@@ -82,9 +82,9 @@ trait Chainable[A, B] {
    * @param labels The labels to use when fitting the LabelEstimator. Must be zippable with the training data.
    */
   final def andThen[C, L](
-    est: LabelEstimator[B, C, L],
-    data: PipelineDatasetOut[A],
-    labels: RDD[L]
+                           est: LabelEstimator[B, C, L],
```
etrain: fix indent?

src/main/scala/workflow/graph/Chainable.scala
```diff
@@ -116,9 +116,9 @@ trait Chainable[A, B] {
    * @param labels The labels to use when fitting the LabelEstimator. Must be zippable with the training data.
    */
   final def andThen[C, L](
-    est: LabelEstimator[B, C, L],
-    data: PipelineDatasetOut[A],
-    labels: PipelineDatasetOut[L]
+                           est: LabelEstimator[B, C, L],
```
etrain: ditto

shivaram:
Thanks @tomerk for the updates. The change looks pretty good to me.

etrain:
Thanks for all the work on this @tomerk !! Looks great - merging this..

src/main/scala/workflow/graph/Rule.scala
etrain: yes. (resolved in a different spot)

src/main/scala/workflow/graph/GraphExecutor.scala
etrain: This is the comment where I realized why the prefixes map needed to be passed forward.

src/main/scala/workflow/graph/EquivalentNodeMergeRule.scala
etrain: Yes, this was resolved.

PR #278, opened 2016-05-29
Samples GMM ref wasn't released in EncEval code

Vaishaal:
I'm pretty sure this is what screwed me up last semester. Sigh.


shivaram:
LGTM. The change looks good, but I'm not sure this can affect things much. The GMM is only run on the driver and only run once ?

Vaishaal:
Oh well. Still worth fixing.

PR #279, opened 2016-06-03
Prefix autocacherule and nodeoptrule

tomerk:
This should add the autocaching rule and node-level optimization rule, bringing the workflow.graph code up to snuff with the workflow code.

### DIFF
--- a/src/main/scala/workflow/graph/AutoCacheRule.scala
+++ b/src/main/scala/workflow/graph/AutoCacheRule.scala
@@ -1,0 +1,696 @@
+package workflow.graph
+
+import breeze.linalg.{DenseMatrix, DenseVector, max}
+import org.apache.spark.storage.StorageLevel
+import org.apache.spark.util.SparkUtilWrapper
+import pipelines.Logging
+import workflow.AutoCacheRule.{AggressiveCache, CachingStrategy, GreedyCache}
+import workflow.WorkflowUtils
+
+case class Profile(ns: Long, rddMem: Long, driverMem: Long) {
+  def +(p: Profile) = Profile(this.ns + p.ns, this.rddMem + p.rddMem, this.driverMem + p.driverMem)
+}
+
+case class SampleProfile(scale: Long, profile: Profile)
+
+class AutoCacheRule(cachingMode: CachingStrategy) extends Rule with Logging {
+
+  /**
+   * Get the operator weights: estimates for how many passes an operator will make over its input dependencies
+   */
+  def getNodeWeights(graph: Graph): Map[NodeId, Int] = {
+    graph.operators.mapValues {
+      case op: WeightedOperator => op.weight
+      case _ => 1
+    }
+  }
+
+  /**
+   * Get all descendents of all sources in the graph
+   */
+  def getDescendantsOfSources(graph: Graph): Set[NodeId] = {
+    graph.sources.foldLeft(Set[GraphId]()) {
+      case (curDescendents, source) => curDescendents ++ AnalysisUtils.getDescendants(graph, source)
+    }.collect {
+      case node: NodeId => node
+    }
+  }
+
+  /**
+   * Get a map representing the children for each node
+   * Note: This doesn't capture how many times each child depended on the instruction
+   */
+  def getChildrenForAllNodes(graph: Graph): Map[NodeId, Seq[GraphId]] = {
+    graph.nodes.map {
+      id => (id, AnalysisUtils.getChildren(graph, id).toSeq)
+    }.toMap
+  }
+
+  /**
+   * Get an estimate for how many times the output of each node will be accessed, assuming
+   * the given set of nodes have their outputs cached.
+   *
+   * Note: This assumes all sinks are accessed exactly once!
+   */
+  def getRuns(
+    linearization: Seq[GraphId],
+    childrenByNode: Map[NodeId, Seq[GraphId]],
+    cached: Set[NodeId],
+    nodeWeights: Map[NodeId, Int]
+  ): Map[NodeId, Int] = {
+
+    linearization.collect {
+      case node: NodeId => node
+    }.foldRight(Map[NodeId, Int]()) {
+      case (node, runsByNode) => {
+        val runs = childrenByNode(node).map {
+          case child: SinkId => 1
+          case child: NodeId => if (cached.contains(child)) {
+            nodeWeights(child)
+          } else {
+            nodeWeights(child) * runsByNode(child)
+          }
+          case _: SourceId => throw new IllegalArgumentException("Invalid graph, node has source as child")
+        }.sum
+
+        runsByNode + (node -> runs)
+      }
+    }
+  }
+
+  /**
+   * Get the initial set of what nodes will have their results effectively cached
+   */
+  def initCacheSet(graph: Graph): Set[NodeId] = {
+    graph.nodes.filter { graph.getOperator(_) match {
+      case DatasetOperator(rdd) => rdd.getStorageLevel != StorageLevel.NONE
+      case _: DatumOperator => true
+      case _: Cacher[_] => true
+      case _: TransformerOperator => false
+      case _: EstimatorOperator => true
+      case _: DelegatingOperator => false
+      case _: ExpressionOperator => true
+    }}
+  }
+
+  /**
+   * This method takes a sequence of profiles at different sample sizes,
+   * and generalizes them to a new data scale by fitting then using linear models
+   * for memory and cpu usage dependent on data scale.
+   */
+  def generalizeProfiles(newScale: Long, sampleProfiles: Seq[SampleProfile]): Profile = {
+    def getModel(inp: Iterable[(Long, String, Long)]): Double => Double = {
+      val observations = inp.toArray
+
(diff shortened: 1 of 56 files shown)


src/main/scala/workflow/graph/LabelEstimator.scala
```diff
@@ -95,6 +95,6 @@ abstract class LabelEstimator[A, B, L] extends EstimatorOperator {
    * @param labels The estimator's training labels
    * @return A new transformer
    */
-  protected def fit(data: RDD[A], labels: RDD[L]): Transformer[A, B]
+  def fit(data: RDD[A], labels: RDD[L]): Transformer[A, B]
```
etrain: We talked about this last night - makes sense.

src/main/scala/workflow/graph/NodeOptimizationRule.scala
```diff
@@ -0,0 +1,200 @@
+package workflow.graph
+
+import org.apache.spark.rdd.RDD
+import workflow.WorkflowUtils
+
+/**
+ * A modified graph executor used by the NodeOptimizationRule, to keep track
```
etrain: I'm a little confused why this is a separate class and why it doesn't extend anything?

Can you describe what this thing is supposed to actually do? Is it for walking through the graph and collecting samples?

etrain: Also, how different is it from the normal GraphExecutor? Looks like there's probably a fair bit of duplicated logic here?

src/main/scala/workflow/graph/NodeOptimizationRule.scala
```diff
@@ -0,0 +1,200 @@
...
+        val labelsDep = executor.getRDD[l](graph.getDependencies(node).apply(1).asInstanceOf[NodeId])._1
+
+        val optimizedOp = op.optimize(dataDep, labelsDep, numPerPartition)
+
+        curGraph.setOperator(node, optimizedOp)
+      }
+    }
+
+    // Unpersist anything the executor cached
+    executor.unpersist()
```
etrain: Ok - this basically looks reasonable to me aside from the GraphExecutor stuff.

src/main/scala/workflow/graph/PipelineEnv.scala
```diff
@@ -30,6 +30,14 @@ class PipelineEnv {
   def setOptimizer(optimizer: Optimizer): Unit = {
     _optimizer = optimizer
   }
+
+  /**
+   * Reset this PipelineEnv (clear state and set the Optimizer to the DefaultOptimizer)
+   */
+  def reset(): Unit = {
+    state.clear()
+    setOptimizer(DefaultOptimizer)
```
etrain: Hmm.. not fully clear why this should change the optimizer?

src/main/scala/workflow/graph/AutoCacheRule.scala
```diff
@@ -0,0 +1,789 @@
...
+    def getModel(inp: Iterable[(Long, String, Long)]): Double => Double = {
+      val observations = inp.toArray
+
+      //Pack a data matrix with observations
+      val X = DenseMatrix.ones[Double](observations.length, 2)
+      observations.zipWithIndex.foreach(o => X(o._2, 0) = o._1._1.toDouble)
+      val y = DenseVector(observations.map(_._3.toDouble))
+      // We solve a linear model for how either memory or time varies with respect to data size
+      // (We train separate models for time and memory usage)
+      val model = max(X \ y, 0.0)
```
etrain: I still am shocked that this works, but it does!

src/main/scala/workflow/graph/AutoCacheRule.scala
```diff
@@ -0,0 +1,789 @@
...
+        val cached = initCacheSet(plan)
+        val nodeWeights = getNodeWeights(plan)
+        val childrenByNode = getChildrenForAllNodes(plan)
+        val linearization = AnalysisUtils.linearize(plan)
+        val runs = getRuns(linearization, childrenByNode, cached, nodeWeights)
+        val descendantsOfSources = getDescendantsOfSources(plan)
+
+        // We have the possibility of caching all uncached nodes accessed more than once,
+        // That don't depend on the sources.
+        // TODO: MUST profile all Cacher nodes!!! (to get cachedMem to be correct)
```
etrain: Has this been done?

etrain:
Can you provide a description of what had to change between the AutoCache rule with the old backend and the current one? A bunch of it I feel like I'm just reading over the same old logic, but there are other bits that are probably new - I guess I'm looking for areas that should be checked with extra scrutiny. 

Additionally - it looks like there is likely some duplicated logic in handling the various types of node in the autocache rule that could be factored out.

Otherwise this is in pretty good shape. Thanks @tomerk !

src/main/scala/workflow/graph/PipelineEnv.scala
tomerk: It resets the Optimizer to the default. This is because some unit tests change the optimizer to test different rules or validate things about the execution, and we eventually have to change it back for other unit tests.

src/main/scala/workflow/graph/NodeOptimizationRule.scala
tomerk: Yes, this class is for walking through the graph and collecting samples for the node-level optimization. I'll rename it to `SampleCollector` and mention that in the docs. The actual process of walking through the DAG is somewhat changed, which is why I didn't make it extend GraphExecutor.

Notable differences : It doesn't optimize, doesn't keep track of prefixes, must sample all RDDs and later unpersist them, and must keep track of numPerPartition at each node.

src/main/scala/workflow/graph/PipelineEnv.scala
etrain: Would a user want this functionality or does this belong in PipelineContext?

On Jun 3, 2016, at 9:02 AM, Tomer Kaftan <notifications@github.com> wrote:

In src/main/scala/workflow/graph/PipelineEnv.scala
<https://github.com/amplab/keystone/pull/279#discussion_r65732447>:

> @@ -30,6 +30,14 @@ class PipelineEnv {
>    def setOptimizer(optimizer: Optimizer): Unit = {
>      _optimizer = optimizer
>    }
> +
> +  /**
> +   * Reset this PipelineEnv (clear state and set the Optimizer to the DefaultOptimizer)
> +   */
> +  def reset(): Unit = {
> +    state.clear()
> +    setOptimizer(DefaultOptimizer)

It resets the Optimizer to the default. This is because some unit tests
change the optimizer to test different rules or validate things about the
execution, and we eventually have to change it back for other unit tests.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/amplab/keystone/pull/279/files/653be30f2fdd1eb649cc69cb23929728c3d74c42#r65732447>,
or mute the thread
<https://github.com/notifications/unsubscribe/ABQ8Ze3p_86Uu21nev8Wa_LbgoBqUS7tks5qIFABgaJpZM4ItH40>
.

tomerk: I had to put this here because state is marked `private[graph]`, so the `PipelineContext` can't access it to reset it.

src/main/scala/workflow/graph/AutoCacheRule.scala
tomerk: No, not yet. I wasn't sure if you wanted to profile all nodes marked cached (to keep track of driver mem), or just the cached rdds to get an accurate cluster mem.

tomerk:
The main shift in AutoCacheRule was going from dealing w/ different forms of instructions to dealing with Operators. The immediate sampling/profiling/profile generalization code for each thing should be about the same as it used to be, but the logic to get to that stage has moved around / possibly changed. One example is that Transformer needs separate logic paths for Datum and RDD now because we can insert Datums directly into the graph.

etrain:
This LGTM! Can you add a workflow.graph version of LeastSquaresEstimator that moves the "densification/sparsification" into the Estimator like we talked about? This would simplify migration of the existing codebase to this backend and ultimately make for easier testing.

Do you want to take a quick pass @shivaram? Most of the changes are actually duplicate logic of stuff we already have, but written now against the new internal pipeline representation. 

I will plan to "flip the switch" early next week and have verified that the previous set of changes actually works on our existing pipelines.

src/main/scala/workflow/graph/AutoCacheRule.scala
etrain: Fine to leave them out, since fitted transformers shouldn't count against cluster memory.

etrain:
Alright, we talked about #280 and the LeastSquaresEstimator offline and have a plan to do this early next week.

src/main/scala/workflow/graph/PipelineEnv.scala
shivaram: We should make this method `private[keystone]` or something like that where its visible to internal classes only (i.e. to avoid users calling this). @etrain you can probably do this in your flip-the-switch PR

shivaram:
Sorry for the delay in looking at this. Thanks @tomerk - LGTM, Merging this

PR #280, opened 2016-06-03
Chain utils

tomerk:


etrain:
Hmm.. in #279 i was specifically referring to making a version of LeastSquaresEstimator where .optimize returns an Estimator rather than a pipeline. This looks like tools to support that, but can you actually do that last step?

etrain:
We talked about this offline. There are some namespace issues that make this hard right now, so I'll incorporate the new LeastSquaresEstimator as part of my "Flip the Switch" PR.

shivaram:
LGTM. Thanks @tomerk -  Merging this

PR #281, opened 2016-06-05
[HOTFIX] Fix broken build by adding ClassTag to ChainUtils

shivaram:
Compiles and tests pass locally

PR #282, opened 2016-06-07
Backend switch

etrain:
This is a pretty substantial PR, designed to "flip the switch" to using the backend developed by @tomerk in #270 #277 #280 #281. 

To provide a high level summary of the changes:

1. package `workflow` was renamed to `workflow.old` (and subsequently removed)
2. `workflow.graph` was renamed to `workflow`
3. `OptimizableEstimators` were changed to reflect the new API.
4. `pipelines` were updated - in many places we returned an RDD and now return a PipelineDataset, and a `.get` is the easiest way around this.
5. Places where we used `.fittedTransformer` are updated to use new `withData` syntax.
6. Things that were marked `private[graph]` are now marked `private[workflow]`
7. `PipelineEnv.reset()` was marked `private[workflow]`, consequently, `pipelines.PipelineContext` (in test) had to be moved to `workflow`, which is likely where it belonged anyway.
8. Unit tests were updated to reflect new package names and new APIs.

The main intent for 1 and 2 is to make it so that user programs only have to make small changes (e.g. dealing with `PipelineDataset` instead of `RDD`.

All the unit tests pass locally, and I verified on a slightly older version of this branch that the newsgroups pipeline runs locally and takes about the same amount of time. I am planning to run integration tests tomorrow before we merge.

Unfortunately github's diff viewer makes it look like there are a lot more changes than there need to be - it might be easiest to look at this commit by commit. For example, there was a file in `src/main/scala/workflow/AutoCacheRule.scala` and another one `src/main/scala/workflow/graph/AutoCacheRule.scala` and the latter has replaced the former, but it looks like we've changed the former inline in this PR, which is not the case (none of the logic has changed other than the package name.)

Feel free to start looking @shivaram - if you have suggestions about how to make the diff viewer behave better I'm open to making changes to these commits.


etrain:
Oh, and in terms of usability, I note three notable regressions

1. PipelineDataset is a bit of a mind-bender. It's a lazy wrapper around a lazy object.
2. Transformers no longer extend Pipeline, which means that we have to sprinkle `toPipeline` in a few places where we'd use them otherwise. (and also unmark `.toPipeline` as `private[workflow]`)
3. Somehow we broke type inference on some of the NLP nodes which handle generic input types. This is evidenced by the changes I had to make to the text pipelines.

I can live with all of these.

src/main/scala/evaluation/AugmentedExamplesEvaluator.scala
```diff
@@ -1,5 +1,7 @@
 package evaluation
 
+import workflow.PipelineDataset
```
shivaram: Do we need an import here if we don't use this class below ?

shivaram:
I see that the diff viewer is trying to show diff between workflow.graph and workflow instead of showing it as a move. 

Here is one thing we could try:  Create one PR which just moves workflow to workflow.old  and then another PR which moves workflow.graph to workflow -- We could do it in a separate branch if we are worried about breaking `master` in between ? If you think that too complicated don't bother, I can review this later today

etrain:
Hmm.. I could do something cherry-picking the commits but the changes to
test happened only after I made (basically) all the changes to main, so
isolating those file movements might be tricky.

I think that looking at the diffs at the level of the individual commits
will give you the same effect as reviewing two individual PRs (most of the
commits are small) - let me know if its too much of a pain and I'll do what
you suggest above.

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
```diff
@@ -99,7 +99,7 @@ object RandomPatchCifarAugmented extends Serializable with Logging {
     val testPredictions = predictionPipeline(testImagesAugmented)
 
     val testEval = AugmentedExamplesEvaluator(
-      testImageIdsAugmented, testPredictions, testLabelsAugmented, numClasses)
+      testImageIdsAugmented, testPredictions.get, testLabelsAugmented, numClasses)
```
shivaram: Minor: But we could make AugmentedExamplesEvaluator take in two PipelineDatasets instead of one PipelineDataset and RDD

src/main/scala/pipelines/images/imagenet/ImageNetSiftLcsFV.scala
```diff
@@ -45,12 +45,12 @@ object ImageNetSiftLcsFV extends Serializable with Logging {
...
-        new BatchPCATransformer(convert(csvread(new File(fname)), Float).t)
+        new BatchPCATransformer(convert(csvread(new File(fname)), Float).t).toPipeline
       case None =>
-        val pca = sampledColumns andThen
-            (ColumnPCAEstimator(numPCADesc), trainingData)
+        val sampler = ColumnSampler(numColSamplesPerImage).toPipeline
+        val pca = ColumnPCAEstimator(numPCADesc) withData (sampler(prefix(trainingData)))
 
-        pca.fittedTransformer
+        pca.toPipeline
```
shivaram: `toPipline` is much better than `fittedTransformer` :)

src/main/scala/pipelines/images/voc/VOCSIFTFisher.scala
```diff
@@ -67,10 +69,9 @@ object VOCSIFTFisher extends Serializable with Logging {
           csvread(new File(conf.gmmWtsFile.get)).toDenseVector)
         pcaFeaturizer andThen FisherVector(gmm)
       case None =>
-        val fisherVector = pcaFeaturizer andThen
-            ColumnSampler(numGMMSamplesPerImage) andThen
-            (GMMFisherVectorEstimator(conf.vocabSize), trainingData)
-        pcaFeaturizer andThen fisherVector.fittedTransformer
+        val sampler = ColumnSampler(numGMMSamplesPerImage).toPipeline
+        val fisherVector = GMMFisherVectorEstimator(conf.vocabSize) withData (sampler(pcaFeaturizer(trainingData)))
+        pcaFeaturizer
```
shivaram: this should be `fisherVector` ?

Also just to keep this consistent can we use `.withData()` instead of the ` withData ` ?

shivaram:
Ok - I think reviewing by commit was much easier. I think I found only one real issue (fisherVector instead of pcaEstimator) in one of the pipelines. The rest of the code looks pretty good. Let me know once you have the results from the integration tests and I'll do one final pass.

src/main/scala/pipelines/images/voc/VOCSIFTFisher.scala
etrain: Very good catch - may also explain some of my integration test woes. Will fix this and re-run.

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
etrain: We actually only have an [RDD,RDD] interface right now. testLabelsAugmented is an RDD in the current pipeline. For the most part our labels end up being RDD while the predictions end up being `PipelineDataset` which is why I have the [PipelineDataset,RDD] interface in the multiclass classifier evaluator. I agree there should be consistency here, though. Maybe there's something we can do at the level of an "evaluator interface" that will obviate the need for duplicating a bunch of boilerplate logic.

src/main/scala/pipelines/images/voc/VOCSIFTFisher.scala
etrain: Fixed.

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmented.scala
etrain: I think I'll create an issue for this rather than addressing in this PR.

src/main/scala/evaluation/AugmentedExamplesEvaluator.scala
etrain: Fixed.

etrain:
Ok - with that fix all the integration tests run in about the right amount of time:
```

shivaram:
Great - the cifar thing might be just due to not convolving twice ?

etrain:
Yep - looks like we're not computing convolutions twice, which is a result of the lazy pipeline tracking. Want to take one more pass @shivaram and then we can merge?

shivaram:
LGTM. Merging this. Will keep an eye on Jenkins after this is merged.

PR #284, opened 2016-06-08
Kernel Generators and Solver

shivaram:
This PR adds kernel generators, a ridge regression solver and a kernel block model to Keystone. It also includes a CIFAR pipeline that shows how it can be used.

At a high level the user-facing design is as follows:
- Users can create a kernel generator object and pass in its params like gamma etc.
- Users can pass in a kernel generator and training data to a solver which will give a kernel model that can be applied to train or test data.

Internally this works as follows:
- `KernelGenerators` are fit during training to bind the trainData as one of the arguments to `K(x, y)`. This step produces a `KernelTransformer`. We need this as the transformer has only one argument usually and it makes the train vs. test distinction clear.
- `KernelTransformers` can be applied to an RDD to generate a `KernelMatrix`. This is a wrapper class lazily populates the kernel matrix and has a block-column API.
- The `KernelMatrix` is used by a linear system solver that just solves `Kx = Y`. Right now this is a part of `KernelRidgeRegression` class but it can be pulled out.

This was originally developed with the help of @stephentu @rolloff and most of the code was ported from #234 written by @Vaishaal.


shivaram:
Note that while the code compiles and the unit test passes, I still haven't run the CIFAR pipeline on a cluster. I'll do that later today / tomm but I wanted to get the design feedback going before that. I also plan to add a CIFAR augmented kernel pipeline, so we can match the numbers in http://arxiv.org/abs/1602.05310

Vaishaal:
Are you going to add the unit tests from the previous PR?

shivaram:
Ah sorry forgot to commit it - I just copied the unit test you wrote. It passes locally

shivaram:
Also one more note is that the kernel classes are templatized so we can handle other inputs. For example the Yelp workload used a `SparseVector` and a linear kernel and we should be able to handle that directly now

Vaishaal:
Couple high level things: 
* We should note somewhere all these exact kernel solvers solve the problem in the dual, whereas our approximations (the only one we have right now is RandomCosine), solve the problem in the primal.

* At least for RBF we should be consistent with gamma, since a gaussian kernel with parameter gamma is approximated with random cosine 1/gamma (I may be missing a constant).

Would it be easy to fit Nystroem into this framework?

src/main/scala/nodes/learning/KernelBlockModel.scala
```diff
@@ -0,0 +1,77 @@
...
+      val testKernelBlock = testKernelMat(blockIdxs.toSeq)
+      val modelBlockBC = in.context.broadcast(model(block))
+
+      // Update predictions
+      var predictionsNew = predictions.zip(testKernelBlock).map { case(pred, testKernelBB) =>
+        pred :+ (testKernelBB * modelBlockBC.value)
+      }
+
+      // If we are checkpointing update our cache
+      if (in.context.getCheckpointDir.isDefined && block % 25 == 24) {
```
etrain: These seem like magic numbers - do we want to make this a configurable parameter?

src/main/scala/nodes/learning/KernelRidgeRegression.scala
```diff
@@ -0,0 +1,240 @@
...
+          val wBlockNew = lhs \ rhs
+          wLocals(l)(block) = wBlockNew
+          val wBlockOldBC = labels.context.broadcast(wBlockOld)
+          val wBlockNewBC = labels.context.broadcast(wBlockNew)
+          (wBlockOldBC, wBlockNewBC)
+        }
+
+        var newModel = updateModel(
+          model, wBlockBCs.map(_._2), blockIdxsBC, preFixLengthsBC).cache()
+        // This is to truncate the lineage every 50 blocks
```
etrain: Is there a reference to a `SPARK-` issue we can put here, or is it just the case that RDDs with long lineage suck.

src/main/scala/nodes/learning/KernelRidgeRegression.scala
```diff
@@ -0,0 +1,240 @@
...
+
+        val wBlockNewBCvalue = wBlockNewBC(idx).value
+        wPart(partInds, ::) := wBlockNewBCvalue(blockInds, ::)
+      }
+      Iterator.single(wParts)
+    }
+    newModel
+  }
+
+  def addMatrixArrays(a: Array[DenseMatrix[Double]], b: Array[DenseMatrix[Double]]) = {
```
stephentu: Should this go in `MatrixUtils` (and be templated as `Seq[DenseMatrix[Double]]`?)

src/main/scala/nodes/learning/KernelRidgeRegression.scala
shivaram: Long lineages are not an issue that have a known solution as far as I know. It will eventually lead to a stack overflow at some point. I wish there was a way to take in a keystone-level global config instead of passing these things through various nodes.

For now I'll just add an optional Int parameter `checkpointNumBlocks` and use it here.

src/main/scala/nodes/learning/KernelBlockModel.scala
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.spark.rdd.RDD
+
+import nodes.stats.{StandardScalerModel, StandardScaler}
+import nodes.util.{VectorSplitter, Identity}
+
+import utils.{MatrixUtils, Stats}
+import workflow.{Transformer, LabelEstimator}
+
+/**
+ * @param models
```
etrain: Can we get some more complete documentation on what this class is?

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -0,0 +1,163 @@
...
+
+  /**
+   * Create a kernel transformer using the given dataset as one of the arguments
+   * to the kernel function. That is, if the kernel function is \phi(x, y), this
+   * binds one of the arguments using the given data.
+   *
+   * @param trainData training data matrix to be used for kernel transformation
+   * @return a kernel transformer that can be applied to train or test data
+   */
+  def fit(trainData: RDD[T]): KernelTransformer[T]
```
etrain: If we made `KernelTransformer extends Transformer[T,DenseVector[Double]]`, and have a `KernelTransformer` also need this special `apply(data: RDD[T]): KernelMatrix` interface (maybe we call it `computeKernelMatrix(data: RDD[T]): KernelMatrix`), then maybe we could build up kernel pipelines with `Features andThen KernelGenerator andThen KernelSolver` as opposed to what we're doing below? 

I'm just trying to fit the kernel stuff into our current APIs rather than inventing new ones if we can help it. I also wonder if we should  be introducing a `KernelSolver` that generally knows how to take `KernelMatrix` as input and emit a `KernelModel` which is a thing that is parameterized by the `KernelTransformer` it uses.

shivaram: The reason we can't do `KernelTransformer extends Transformer[T, DenseVector[Double]` is because we can't return a `RDD[DenseVector[Double]]` as an output from `apply(RDD[T])`. This is a limitation of our Transformer API which enforces the pattern to be row-oriented while the `KernelMatrix` here is column block oriented.  Or to put it another way the trait definition where `T => U` and `RDD[T] => RDD[U]` means that we can't use custom datastructures like KernelMatrix in the type definition.

src/main/scala/nodes/learning/KernelRidgeRegression.scala
shivaram: Done.

src/main/scala/nodes/learning/KernelBlockModel.scala
shivaram: Done

shivaram:
@Vaishaal 
1. Good point. I added a comment in the class documentation for KRR that we solve this in the dual.

2. The RBF definition we are using seems pretty standard from what I see elsewhere http://scikit-learn.org/stable/modules/metrics.html#rbf-kernel - Is Figure 1 from http://people.eecs.berkeley.edu/~brecht/papers/07.rah.rec.nips.pdf the best definition of the mapping ? We can make a note of this somewhere in the docs.

3. I think Nystrom solver should fit pretty well in this framework. It can still take in a `KernelMatrix` as the input and then just sample the necessary column blocks. The main problem is that we might need to write another model class that does the same sampling on the test blocks as well. I'll try to see if we can make this more general though

Vaishaal:
@stephentu had some comments about this. I think its weird cause the Nystroem is solved in the primal IIRC? I might be wrong..

And yes that figure is the best mapping.

shivaram:
Yeah Nystrom is solved in the primal. But the `KernelMatrix`, which is this lazy intermediate datastructure, should remain the same across primal or dual solvers. 

Also ideally we should be able to use the same block linear mapper that we use for RandomCosines but since the kernel transformer is not a keystone transformer we need this other class -- and this is something we can try to generalize better.

src/main/scala/nodes/learning/KernelGenerator.scala
etrain: Hmm.. I guess what I'm saying is that `KernelMatrix` could have both a column block access method *and* a Row access method. Certainly there's nothing that says a KernelMatrix is a row-partitioned matrix or a column partitioned matrix - but our solver currently needs the column blocked access pattern and so we need a way of telling the system "hey - for this part we want a `Seq[RDD[BDV[Double]]]`"  What's unclear to me is how having both interfaces would play with internal fitting code. I think we'd need to write a special case for this and it might not be worth it. 

Let me see if I can come up with a clean way to handle this over the weekend, otherwise we can proceed and merge without this on monday.

shivaram:
@etrain This is worth another look when you get a chance. From testing it on a single machine, I can get the same test error (around 20%) as the older code for CIFAR unaugmented.  In terms of performance I've made a bunch of changes that brings the performance pretty close to what we had before.  

However the kernel matrix API is now a bit more tricky to use (users need to call unpersist after they are done with a block). I think this is a reasonable trade-off given that these classes are internal to keystone and the user-facing API is still straightforward to use. But any ideas to improve the API are welcome.

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -66,8 +66,19 @@ trait KernelTransformer[T] {
    */
   def apply(data: T): DenseVector[Double]
 
-  // Internal function used to lazily populate the kernel matrix.
-  private[learning] def computeKernel(data: RDD[T], idxs: Seq[Int]): RDD[DenseVector[Double]]
+  /**
+   * Internal function used to lazily populate the kernel matrix.
+   *
+   * NOTE: This function returns a *cached* RDD and the caller is responsible
```
etrain: since this is marked `private[learning]` I'm ok with this.

etrain:
This looks pretty good to me. I don't see any major changes I'd make. The state management stuff is annoying but manageable. We could think of a standard interface that lets things clean up after themselves. This KernelGenerator stuff is a weird corner case, though, and I don't know if this is the right place to start with it.

src/main/scala/nodes/learning/KernelGenerator.scala
```diff
@@ -154,14 +158,18 @@ class GaussianKernelTransformer(
...
-        val term1 = xxt(idx, ::).t * (-2.0)
-        val term2 = DenseVector.fill(xxt.cols)(dataDotProdVal)
-        val term3 = trainBlockDotProdBC.value
-        val term4 = (term1 + term2 + term3) * (-gamma)
-        exp(term4)
+      if (iterXXT.hasNext) {
+        val xxt = iterXXT.next()
+        assert(iterXXT.isEmpty)
+        iterDataDotProds.zipWithIndex.map { case (dataDotProdVal, idx) =>
+          val term1 = xxt(idx, ::).t * (-2.0)
```
etrain: is there a more descriptive (math) name for these variables? If not its fine.

src/main/scala/pipelines/images/cifar/RandomPatchCifarAugmentedKernel.scala
```diff
@@ -171,6 +173,7 @@ object RandomPatchCifarAugmentedKernel extends Serializable with Logging {
     // NOTE: ONLY APPLICABLE IF YOU CAN DONE COPY-DIR
     conf.remove("spark.jars")
     val sc = new SparkContext(conf)
+    appConfig.checkpointDir.foreach(dir => sc.setCheckpointDir(dir))
```
etrain: monads ftw.

src/main/scala/nodes/learning/KernelGenerator.scala
shivaram: I added a comment on what was going on above this. Let me know if that helps

etrain: Yeah, the new comments look good to me - thanks!

etrain:
Alright - finally merging this! Thanks @shivaram @Vaishaal @stephentu @rolloff  - this is a great new feature.

PR #286, opened 2016-06-15
Exclude slf4j-simple from Scala NLP deps

shivaram:
Not excluding this means that our `log4j.properties` gets ignored as SLF4J uses `slf4j-simple` as the logging backend. To test this you can run `sbt/sbt test` without this change and you should see lines like

```
[Executor task launch worker-0] INFO org.apache.spark.storage.BlockManager - Found block rdd_150_1 locally
```
which will no longer appear with this change


etrain:
Nice digging to get this back to normal. Thanks @shivaram ! LGTM

PR #292, opened 2017-02-24
Evaluatorinterface

### LINKED ISSUE #283: Create "Evaluator" interface.
Generally, our evaluators are pretty simple - they take in two datasets of a particular type (predictions, labels) and output some score. They are parameterized differently (e.g. num classes, etc.) but this could be a constructor parameter.

With the advent of `PipelineDataset`, doing the "right thing" based on inputs being dataset/dataset, rdd/rdd, or a mix of both is boilerplate logic that we should only write down once. 

I propose a `trait Evaluator` that looks something like this:

```
trait Evaluator[P,L,E] {
  def apply(predictions: RDD[P], labels: RDD[L]): E
  def apply(predictions: PipelineDataset[P], labels: RDD[L]) = apply(predictions.get, labels)
  def apply(predictions: RDD[P], labels: PipelineDataset[L]) = apply(predictions, labels.get)
  def apply(predictions: PipelineDataset[P], labels: PipelineDataset[L]) = apply(predictions.get, labels.get)
}
```

Other options include doing something clever with implicits, but I feel like this is more explicit.

etrain:
Closes #283.

### DIFF
--- a/src/main/scala/evaluation/Evaluator.scala
+++ b/src/main/scala/evaluation/Evaluator.scala
@@ -4,8 +4,8 @@ import org.apache.spark.rdd.RDD
 import workflow.PipelineDataset
 
 trait Evaluator[P,L,E] {
-  def apply(predictions: RDD[P], labels: RDD[L]): E
-  def apply(predictions: PipelineDataset[P], labels: RDD[L]): E = apply(predictions.get, labels)
-  def apply(predictions: RDD[P], labels: PipelineDataset[L]): E = apply(predictions, labels.get)
-  def apply(predictions: PipelineDataset[P], labels: PipelineDataset[L]): E = apply(predictions.get, labels.get)
+  def evaluate(predictions: RDD[P], labels: RDD[L]): E
+  def evaluate(predictions: PipelineDataset[P], labels: RDD[L]): E = evaluate(predictions.get, labels)
+  def evaluate(predictions: RDD[P], labels: PipelineDataset[L]): E = evaluate(predictions, labels.get)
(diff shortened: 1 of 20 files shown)


src/main/scala/evaluation/Evaluator.scala
```diff
@@ -0,0 +1,11 @@
+package evaluation
+
+import org.apache.spark.rdd.RDD
+import workflow.PipelineDataset
+
+trait Evaluator[P,L,E] {
```
tomerk: I actually posted this comment on friday, but I forgot to press 'submit review' so I guess it didn't post until now...

etrain: Fixed.

shivaram: Might want to add scaladoc here ?

src/main/scala/evaluation/MeanAveragePrecisionEvaluator.scala
```diff
@@ -7,23 +7,24 @@ import org.apache.spark.SparkContext._
...
   /**
    * Compute the mean average precision for multi-class classification
    * NOTE: The current implementation is only suitable when we have a 
    *       small number of classes and data items.
-   * @param testActual - For every test image, this contains list of valid labels. 
+   * @param testActual For every test image, this contains list of valid labels.
    *                     Labels are assumed to be class ids.
-   * @param testPredicted - For every test image, this contains a list of scores for each class
-   * @param numClasses - Total number of classes
+   * @param testPredicted For every test image, this contains a list of scores for each class
```
shivaram: nit - swap the order of doc here to match argument order ?

PR #295, opened 2017-03-02
Add example pipeline scripts.

### LINKED ISSUE #289: Create and populate scripts to run example pipelines
While we have a number of pipelines checked into `pipelines/` but dont have scripts which select the right hyper parameters to run them. We should add scripts to say `examples/`  and also use those scripts to populate http://keystone-ml.org/examples

etrain:
Fixes #289 

This PR adds three example scripts that will actually download data and run pipelines for four of the example pipelines that ship with keystone.

### DIFF
--- a/examples/images/cifar_random_patch.sh
+++ b/examples/images/cifar_random_patch.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+set -e
+
+#Set environment variables
+: ${KEYSTONE_MEM:=4g}
+export KEYSTONE_MEM
+
+KEYSTONE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"/../..
+: ${EXAMPLE_DATA_DIR:=$KEYSTONE_DIR/example_data}
+
+if [ ! -d $EXAMPLE_DATA_DIR ]; then
+    mkdir $EXAMPLE_DATA_DIR
+fi
+
+#Download data if necessary.
+if [[ ! ( -f $EXAMPLE_DATA_DIR/cifar_train.bin && -f $EXAMPLE_DATA_DIR/cifar_test.bin ) ]]; then
+    #Get the data
+    wget -O $TMPDIR/cifar-10-binary.tar.gz  http://www.cs.toronto.edu/~kriz/cifar-10-binary.tar.gz 
+
+    #Decompress it
+    tar zxvf $TMPDIR/cifar-10-binary.tar.gz -C $TMPDIR
+    cat $TMPDIR/cifar-10-batches-bin/data_batch*.bin > $EXAMPLE_DATA_DIR/cifar_train.bin
+    mv $TMPDIR/cifar-10-batches-bin/test_batch.bin $EXAMPLE_DATA_DIR/cifar_test.bin
+       
+    #Clean up. 
+    rm -rf $TMPDIR/cifar-10-batches-bin
+    rm -rf $TMPDIR/cifar-10-binary.tar.gz
+fi
+
+#Run the pipeline
+$KEYSTONE_DIR/bin/run-pipeline.sh \
+  keystoneml.pipelines.images.cifar.RandomPatchCifar \
+  --trainLocation $EXAMPLE_DATA_DIR/cifar_train.bin \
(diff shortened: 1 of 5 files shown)


examples/images/cifar_random_patch.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+set -e
+
+#Set environment variables
+: ${KEYSTONE_MEM:=4g}
+export KEYSTONE_MEM
+
+KEYSTONE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"/../..
```
tomerk: How robust is this command for getting the Keystone directory?

examples/images/mnist_random_fft.sh
```diff
@@ -0,0 +1,34 @@
...
+: ${EXAMPLE_DATA_DIR:=$KEYSTONE_DIR/example_data}
+
+if [ ! -d $EXAMPLE_DATA_DIR ]; then
+    mkdir $EXAMPLE_DATA_DIR
+fi
+
+# Get the data from S3
+if [ ! -f $EXAMPLE_DATA_DIR/train-mnist-dense-with-labels.data ]; then
+    wget -O $EXAMPLE_DATA_DIR/train-mnist-dense-with-labels.data  \
+        http://mnist-data.s3.amazonaws.com/train-mnist-dense-with-labels.data
```
tomerk: Just to confirm: These s3 buckets are public access?

examples/images/voc_sift_fisher.sh
```diff
@@ -0,0 +1,31 @@
...
+export KEYSTONE_MEM
+
+KEYSTONE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"/../..
+: ${HADOOP_EXAMPLE_DATA_DIR:=/example_data}
+
+
+#Get the data and copy to HDFS if necessary.
+if [ ! hadoop fs -test -e $HADOOP_EXAMPLE_DATA_DIR/VOCtrainval_06-Nov-2007.tar ]; then
+    wget -O $TMPDIR/VOCtrainval_06-Nov-2007.tar https://s3-us-west-2.amazonaws.com/voc-data/VOCtrainval_06-Nov-2007.tar
+    hadoop fs -copyFromLocal VOCtrainval_06-Nov-2007.tar $HADOOP_EXAMPLE_DATA_DIR
```
tomerk: Is this the only one of these example scripts we expect to be run on a cluster?

etrain: Yeah, for now. The TIMIT and Amazon datasets require special agreements to redistribute, so we can't just make them publicly available. Honestly the input data for VOC is small enough that it might be fine to refactor such that it's on the master and then `sc.parallelize()` the training data.

examples/images/mnist_random_fft.sh
etrain: Yes.

examples/images/cifar_random_patch.sh
etrain: I dunno, @shivaram is the bash ninja. This is the top answer on stack overflow for "current bash script directory." :)

examples/images/voc_sift_fisher.sh
etrain: We could also have an ImageNet thing since that's public, such a script would look like this though I hesitate to include it because to run in a reasonable amount of time you need n>10 machines.

tomerk: Makes sense. I'm assuming we're going to add some docs somewhere about which of these examples run locally vs on a cluster?

etrain: I have updated the example to not use hadoop and it will use the local filesystem and verified that it runs locally (if you give it enough memory). The way the loader is written it will work fine with data that's on the local filesystem as well as on HDFS, and given the size of the data it won't slow things down dramatically to read off local disk and then shuffle over the network.

etrain:
Unless you have objections @tomerk I think we can go ahead and merge this. Have verified that these scripts do what I expect locally on MacOS and Linux.

tomerk:
Seems fine to me!

